PR #16, opened 2023-09-06
feat: add render Users, Ways and DayReport table on mainPage

Ekaterina1994:


mw-webapp/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Formatted date
+ */
+export class DateUtils {
+
+  /**
+   * Get formatted date in ISO format yyyy-mm-dd
```
tritonJS826, line 7:
```suggestion
   * Format date to string yyyy-mm-dd
```

tritonJS826 (approved):

SergioAJS (approved):

PR #32, opened 2023-09-09
#21 feat: add render table's cells with correct info from db

Ekaterina1994:


mw-webapp/src/model/firebaseCollection/time/Time.ts
```diff
@@ -0,0 +1,31 @@
+import {TimeUnit} from "src/model/businessModel/time/timeUnit/TimeUnit";
+
+/**
+ * What was done
```
tritonJS826, line 4: What)

mw-webapp/src/model/firebaseCollection/time/Time.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export class Time {
+
+  /**
+   * Unit of time measurement
+   */
+  public timeUnit: TimeUnit;
+
+  /**
+   * Number of time units (minutes) spent doing the job
```
tritonJS826, line 14: Not minutes

tritonJS826 (approved):

SergioAJS (approved):

mw-webapp/src/model/firebaseCollection/time/Time.ts
Ekaterina1994: Change comment to "How long was the job done"

mw-webapp/src/model/firebaseCollection/time/Time.ts
Ekaterina1994: Deleted "minutes"

PR #35, opened 2023-09-10
#20, #23 feat: redirect after auth and logout button

SergioAJS:


mw-webapp/src/component/usersBlock/UsersBlock.tsx
```diff
@@ -17,7 +17,7 @@ export const UsersBlock = () => {
   }, []);
 
   const renderUsers = () => {
-    return users.map((user) => (
+    return Object.values(users).map((user) => (
```
Ekaterina1994, line 20: Better create const instead of Object.values....

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,62 @@
-import {Link} from "react-router-dom";
+import {User, getRedirectResult, onAuthStateChanged} from "firebase/auth";
+import {useEffect, useState} from "react";
+import {Link, useNavigate} from "react-router-dom";
 import {useAuth} from "src/utils/useAuth";
 
 export const AuthPage = () => {
-  const handleGoogleSignIn = useAuth();
+  const {handleGoogleSignIn, handleLogout, writeNewUserData, auth} = useAuth();
```
Ekaterina1994, line 7: Is it possible to rename this consts or it's naming from firebase? For make it more similar - for example to change handleGoogleSignIn -> handleLogIn?

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,62 @@
...
+        setUser(currentUser);
+      } else {
+        setUser(null);
+      }
+    });
+    return () => {
+      handleAuthState();
+    };
+  }, []);
+
```
Ekaterina1994, line 23: I think better to move out the function and then just call it inside useEffect

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,62 @@
...
+        if (error instanceof Error) {
+          errorMessage = error.message;
+        }
+        alert(errorMessage);
+      }
+    };
+    return () => {
+      getNewUserCredentials();
+    };
+  }, []);
```
Ekaterina1994, line 48: Again if possible bette rmove out function. And call it inside useEffect. Also no eed to use 2 useEffect. You can call few functions in 1 useEffect

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,62 @@
...
   return (
     <>
-      <button onClick={handleGoogleSignIn}>
-        Sign in with Google
-      </button>
+      {user ? (
+        <button onClick={handleLogout}>
+          Logout
+        </button>
+      ) : (
```
Ekaterina1994, line 55: Need to use component button, not new buttons. I think that task #23 will be ready after that.

Ekaterina1994, line 55: Need to use component button, not new buttons. I think that task #23 will be ready after that.

mw-webapp/src/utils/useAuth.ts
```diff
@@ -1,17 +1,35 @@
-import {GoogleAuthProvider, getAuth, signInWithRedirect, getRedirectResult} from "firebase/auth";
+import {GoogleAuthProvider, getAuth, signInWithRedirect, signOut} from "firebase/auth";
+import {set, ref} from "firebase/database";
+import {db} from "src/firebase";
 
 export const useAuth = () => {
   const provider = new GoogleAuthProvider();
   const auth = getAuth();
 
+  const writeNewUserData = (userId: string, email: string | null, name: string | null) => {
```
Ekaterina1994, line 9: Better create an interface UserDataProps with these properties

mw-webapp/src/utils/useAuth.ts
```diff
@@ -20,5 +38,6 @@ export const useAuth = () => {
       alert(errorMessage);
     }
   };
-  return handleGoogleSignIn;
+
+  return {handleGoogleSignIn, handleLogout, writeNewUserData, auth};
 };
```
Ekaterina1994, line 43: Also maybe better moveout these functions to keep single responsibility

Ekaterina1994 (changes requested):

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

mw-webapp/src/utils/useAuth.ts
SergioAJS: Done

mw-webapp/src/utils/useAuth.ts
SergioAJS: After that, the existence of the useAuth hook became senseless and I deleted it.
Done

mw-webapp/src/component/usersBlock/UsersBlock.tsx
```diff
@@ -17,7 +17,8 @@ export const UsersBlock = () => {
   }, []);
 
   const renderUsers = () => {
-    return users.map((user) => (
+    const usersObjectToArray = Object.values(users);
```
tritonJS826, line 20: it is not a converter. It is `usersList`

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,36 @@
...
 
 export const AuthPage = () => {
-  const handleGoogleSignIn = useAuth();
+  const [user, setUser] = useState<User | null>(null);
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    return () => {
+      handleUserAuthState(setUser);
+      getNewUserCredentials(navigate, "/main");
```
tritonJS826, line 17: why does getNewUserCredentials returns void, and its interface include navigate and part of url as a string? Fix all this please)

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,36 @@
...
+import {handleLogOut} from "src/utils/auth/handleLogOut";
 
 export const AuthPage = () => {
-  const handleGoogleSignIn = useAuth();
+  const [user, setUser] = useState<User | null>(null);
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    return () => {
+      handleUserAuthState(setUser);
```
tritonJS826, line 16: this callback wil be invoked after unmounting  Auth page. Why do you need to handle auth state in this case and what does it mean. Use more approproate abstraction or logic.

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,14 +1,36 @@
...
+      getNewUserCredentials(navigate, "/main");
+    };
+  }, []);
 
   return (
     <>
-      <button onClick={handleGoogleSignIn}>
-        Sign in with Google
-      </button>
+      {user ? (
```
tritonJS826, line 23: dont use ternarian operator in multiline expressions if possibe. Maybe in this case it will be more appropriate to render the same buttom with different props depeds on does user exist

mw-webapp/src/utils/auth/getNewUserCredentials.ts
```diff
@@ -0,0 +1,24 @@
+import {getRedirectResult} from "firebase/auth";
+import {NavigateFunction} from "react-router-dom";
+import {writeNewUserData} from "src/utils/auth/writeNewUserData";
+import {auth} from "src/firebase";
+
+export const getNewUserCredentials = async (navigateFunction: NavigateFunction, navigateDestination: string) => {
```
tritonJS826, line 6: this function violates single responsibility and it does not get anything. Refactor required

mw-webapp/src/utils/auth/getNewUserCredentials.ts
```diff
@@ -0,0 +1,24 @@
...
+      navigateFunction(navigateDestination);
+    } else {
+      return;
+    }
+  } catch (error) {
+    let errorMessage;
+    if (error instanceof Error) {
+      errorMessage = error.message;
+    }
+    alert(errorMessage);
```
tritonJS826, line 22: why can't you use something simplier like `??` operator or add helper?

mw-webapp/src/utils/auth/handleAuthState.ts
```diff
@@ -0,0 +1,12 @@
+import {User, onAuthStateChanged} from "firebase/auth";
+import {auth} from "src/firebase";
+
+export const handleUserAuthState = (setUserFunction: React.Dispatch<React.SetStateAction<User | null>>) => {
+  onAuthStateChanged(auth, (currentUser) => {
+    if (currentUser) {
+      setUserFunction(currentUser);
+    } else {
+      setUserFunction(null);
+    }
```
tritonJS826, line 10: the same as setUserFunction(currentUser ?? null),

mw-webapp/src/utils/auth/handleLogIn.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const handleLogIn = async () => {
+  try {
+    await signInWithRedirect(auth, provider);
+  } catch (error) {
+    let errorMessage;
+    if (error instanceof Error) {
+      errorMessage = error.message;
+    }
+    alert(errorMessage);
```
tritonJS826, line 12: move to helper or use simplier method like '??' operator

mw-webapp/src/component/usersBlock/UsersBlock.tsx
```diff
@@ -17,7 +17,8 @@ export const UsersBlock = () => {
...
+    return usersObjectToArray.map((user) => (
       <UserCard key={user.uuid}
```
Ekaterina1994, line 22: Please think about naming. Inside function renderUsers we work with users so I think better use const users for map it. You can change naming for state const

mw-webapp/src/utils/auth/handleLogOut.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const handleLogOut = async () => {
+  try {
+    await signOut(auth);
+  } catch (error) {
+    let errorMessage;
+    if (error instanceof Error) {
+      errorMessage = error.message;
+    }
+    alert(errorMessage);
```
tritonJS826, line 12: helper or simplier

mw-webapp/src/component/usersBlock/UsersBlock.tsx
Ekaterina1994, line 22: Please think about naming. Inside function renderUsers we work with users so I think better use const users for map it. You can change naming for state const

mw-webapp/src/utils/auth/handleLogOut.ts
```diff
@@ -0,0 +1,14 @@
+import {signOut} from "firebase/auth";
+import {auth} from "src/firebase";
+
+export const handleLogOut = async () => {
+  try {
+    await signOut(auth);
```
tritonJS826, line 6: what heppend for user if sign out throw error? Nothing except alert? Is it a good scenario?

mw-webapp/src/utils/auth/writeNewUserData.ts
```diff
@@ -0,0 +1,17 @@
...
+import {db} from "src/firebase";
+
+interface UserData {
+  userId: string
+  userEmail: string | null
+  userName: string | null
+}
+
+export const writeNewUserData = ({userId, userEmail, userName}: UserData) => {
+  const usersListRef = ref(db, "/users/" + userId);
```
tritonJS826, line 11: it is very hard to managed widespreaded strings in enterprise projects. move to service

tritonJS826 (changes requested):

tritonJS826:
Call me in the telegram when you will solve all issues and will need my approve (after Kates approve, please)

mw-webapp/src/utils/auth/writeNewUserData.ts
```diff
@@ -0,0 +1,17 @@
+import {ref, set} from "firebase/database";
+import {db} from "src/firebase";
+
+interface UserData {
+  userId: string
+  userEmail: string | null
+  userName: string | null
+}
+
```
Ekaterina1994, line 9: Check syntax. Where is ;

mw-webapp/src/utils/auth/writeNewUserData.ts
```diff
@@ -0,0 +1,17 @@
...
+export const writeNewUserData = ({userId, userEmail, userName}: UserData) => {
```
Ekaterina1994, line 10: Is it possible don't use destructure? If image that user has 100 properties it will be hard to write it all

Ekaterina1994, line 10: maybe this const better rename to createNewUser/ addNewUser? Or something like this? What do you think?

mw-webapp/src/utils/auth/handleAuthState.ts
```diff
@@ -0,0 +1,12 @@
+import {User, onAuthStateChanged} from "firebase/auth";
+import {auth} from "src/firebase";
+
+export const handleUserAuthState = (setUserFunction: React.Dispatch<React.SetStateAction<User | null>>) => {
```
Ekaterina1994, line 4: Check naming. Function should have the same name as file

Ekaterina1994 (changes requested):

mw-webapp/src/component/usersBlock/UsersBlock.tsx
SergioAJS: Done

mw-webapp/src/component/usersBlock/UsersBlock.tsx
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

mw-webapp/src/utils/auth/handleAuthState.ts
SergioAJS: Done, thanks

mw-webapp/src/utils/auth/handleAuthState.ts
SergioAJS: Done

mw-webapp/src/utils/auth/handleAuthState.ts
```diff
@@ -0,0 +1,12 @@
...
+  onAuthStateChanged(auth, (currentUser) => {
+    if (currentUser) {
+      setUserFunction(currentUser);
```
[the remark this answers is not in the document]
SergioAJS: Done

mw-webapp/src/utils/auth/writeNewUserData.ts
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: This function writes user data to database.
It includes navigate to redirect to the page with a table only if the user logged in. If the user already logged in three is no redirection to the page with the table. I write in such way because we don't have logout button on the table page.
Npw I delete navigate params from this function.

mw-webapp/src/utils/auth/getNewUserCredentials.ts
SergioAJS: Done

mw-webapp/src/utils/auth/writeNewUserData.ts
SergioAJS: I remove destructing if you mean it

mw-webapp/src/utils/auth/getNewUserCredentials.ts
SergioAJS: Done, if understand you right

mw-webapp/src/utils/auth/handleLogIn.ts
SergioAJS: Done

mw-webapp/src/utils/auth/handleLogOut.ts
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
Ekaterina1994: Victor mean that if you call function get... then this function should return something so if this function doesn't return anything  need to change naming

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,17 +1,37 @@
...
 export const AuthPage = () => {
-  const handleGoogleSignIn = useAuth();
+  const [user, setUser] = useState<User | null>(null);
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    handleUserAuthState(setUser);
+    return () => {
+      getNewUserCredentials();
+    };
```
Ekaterina1994, line 18: Do you need to use return getNewUserCredentials or maybe just call it will be enough

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,17 +1,37 @@
...
-        Sign in with Google
-      </button>
-      <Link to={"main"}>
-        Workflow
-      </Link>
+      {user ?
+        navigate("/main") : (
+          <>
+            <Button
+              value={user ? "Logout" : "Login"}
```
Ekaterina1994, line 27: Move out text of buttons to consts

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,17 +1,37 @@
...
   return (
     <>
-      <button onClick={handleGoogleSignIn}>
-        Sign in with Google
-      </button>
-      <Link to={"main"}>
-        Workflow
-      </Link>
+      {user ?
+        navigate("/main") : (
```
Ekaterina1994, line 24: create consts with paths. Think maybe not in this file

mw-webapp/src/utils/auth/getNewUserCredentials.ts
```diff
@@ -0,0 +1,18 @@
+import {getRedirectResult} from "firebase/auth";
+import {writeNewUserData} from "src/utils/auth/writeNewUserData";
+import {auth} from "src/firebase";
+
+export const getNewUserCredentials = async () => {
```
Ekaterina1994, line 5: change naming . get.. always return values

Ekaterina1994 (changes requested):

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Now it's enough. Early this function doesn't write user to DB if I call it in useEffect.
Done

mw-webapp/src/utils/auth/getNewUserCredentials.ts
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: When we need to use path consts somewhere else we will create separate file.
For now, I create this const in the same file

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Done

Ekaterina1994 (approved):

mw-webapp/src/component/usersBlock/UsersBlock.tsx
```diff
@@ -17,6 +17,7 @@ export const UsersBlock = () => {
   }, []);
 
   const renderUsers = () => {
+    const users = Object.values(usersList);
```
tritonJS826, line 20: what does this line do? you are trying to get array from array with the same values or i miss siimething?

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -1,17 +1,38 @@
...
+import {Button} from "src/component/button/Button";
+import {handleUserCredentials} from "src/utils/auth/handleUserCredentials";
+import {handleUserAuthState} from "src/utils/auth/handleUserAuthState";
+import {handleLogIn} from "src/utils/auth/handleLogIn";
+import {handleLogOut} from "src/utils/auth/handleLogOut";
+
+const BUTTON_LOG_IN_VALUE = "Login";
+const BUTTON_LOG_OUT_VALUE = "Logout";
+const LINK_TEXT = "Workflow";
+const MAIN_PATH = "/main";
```
tritonJS826, line 13: we have pages.tsx file with router related data. remember about encapsulation.

mw-webapp/src/utils/auth/handleLogIn.ts
```diff
@@ -0,0 +1,10 @@
+import {signInWithRedirect} from "firebase/auth";
+import {auth, provider} from "src/firebase";
+
+export const handleLogIn = async () => {
+  try {
+    await signInWithRedirect(auth, provider);
```
tritonJS826, line 6: should it be in service layer? what heppend for user sign in throw error? Nothing except alert? Is it a good scenario?

mw-webapp/src/utils/auth/handleUserCredentials.ts
```diff
@@ -0,0 +1,18 @@
+import {getRedirectResult} from "firebase/auth";
+import {writeNewUserData} from "src/utils/auth/writeNewUserData";
+import {auth} from "src/firebase";
+
+export const handleUserCredentials = async () => {
+  try {
+    const userCredentials = await getRedirectResult(auth);
```
tritonJS826, line 7: i dont understand why network related methods are in the utils

tritonJS826: @Ekaterina1994

tritonJS826 (approved):

mw-webapp/src/utils/auth/handleLogOut.ts
Ekaterina1994: Не менял поведение при появлении ошибки при Logout, потому что не понял, что еще надо сделать? Редирект на страницу авторизации?
Не уверен, что правильно написал обработчик ошибок с использованием ??

mw-webapp/src/component/usersBlock/UsersBlock.tsx
Ekaterina1994: It''s because of on fb collections is objects with key type string. And when we mapping this data we need firstly convert it to array where key will be 0,1,2... If you know how to map objects with any keys tell us

SergioAJS: If the keys in the database are sequential - 0, 1, 2...n, then we get an array at once and there are no errors, otherwise - an object, and on the object map does not work.
At first, I thought that the error was due to the fact that the key I use as userId contains letters. So I tried first removing all letters and leaving only numbers before using userId as a key. But that didn't work because the resulting number was large and still returned an object
Then I tried to get the length of the users collection from the database to assign the key based on the number of users. This didn't work because I couldn't get the length of the collection before writing a new user to the database
In the end, I decided to use Object.values to get an array of user data.

mw-webapp/src/utils/auth/handleLogIn.ts
SergioAJS: Didn't change the behavior when Logout error appears because I didn't understand what else needs to be done? Redirect to the authorization page?

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: I will move it to the appropriate folder if you tell me where it should be

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Use path from pages.tsx file
By the way, the paths there are specified as strings, not as constants

PR #45, opened 2023-09-13
refactor: #40 add method to UserService that writes new user data

SergioAJS:


mw-webapp/src/service/UserService.ts
```diff
@@ -13,4 +13,16 @@ export class UserService {
     });
   }
 
+  public static writeNewUserData(data: UserDTO) {
+    const usersListRef = ref(db, "/users/" + data.uuid);
+    set(usersListRef, {
+      uuid: data.uuid,
+      email: data.name,
+      name: data.email,
```
tritonJS826, line 21: why name is email

mw-webapp/src/utils/auth/handleUserCredentials.ts
```diff
@@ -1,16 +1,29 @@
 import {getRedirectResult} from "firebase/auth";
-import {writeNewUserData} from "src/utils/auth/writeNewUserData";
 import {auth} from "src/firebase";
+import {UserService} from "src/service/UserService";
+
+const newUserBlankWays = {
+  ownWays: [""],
+  favoriteWays: [""],
+  mentoringWays: [""],
```
tritonJS826, line 8: why do we need it?

mw-webapp/src/utils/auth/handleUserCredentials.ts
```diff
@@ -1,16 +1,29 @@
...
   try {
     const userCredentials = await getRedirectResult(auth);
-    if (userCredentials) {
-      writeNewUserData(
-        {userId: userCredentials.user.uid, userEmail: userCredentials.user.email, userName: userCredentials.user.displayName},
-      );
-    } else {
-      return;
+    if(userCredentials) {
+      if (userCredentials.user.email && userCredentials.user.displayName) {
```
tritonJS826, line 15: 1. nested if - could we avoid it?
2. handleUserCredentials - could we find a little bit more appropriate name, I dont understand what this function should do
3. I dont understand also when I should call this function

mw-webapp/src/utils/auth/handleUserCredentials.ts
```diff
@@ -1,16 +1,29 @@
...
+        UserService.writeNewUserData(
+          {
+            uuid: userCredentials.user.uid,
```
Ekaterina1994, line 18: Is it a special using user.uid (instead of user.uuid)?

mw-webapp/src/utils/auth/handleUserCredentials.ts
```diff
@@ -1,16 +1,29 @@
...
+            email: userCredentials.user.email,
+            name: userCredentials.user.displayName,
```
Ekaterina1994, line 20: Cna't we use name: userCredentials.user.name?

Ekaterina1994 (changes requested):

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: To satisfy User model

mw-webapp/src/service/UserService.ts
SergioAJS: Made changes. Done

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: No, we can't

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: Rename and delete nested if

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: uid, email, displayName - these are the names of properties specified by Firebase

mw-webapp/src/utils/auth/handleUserCredentials.ts
SergioAJS: If you write user.uuid you get a tautology - user.user unique id

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -17,7 +17,7 @@ export const AuthPage = () => {
 
   useEffect(() => {
     handleUserAuthState(setUser);
-    handleUserCredentials();
+    writeNewUserCredentialsToDataBase();
```
Ekaterina1994, line 20: You can write it shorter  - writeNewCredentialsToDb()

Ekaterina1994 (approved):

mw-webapp/src/pages/authPage/AuthPage.tsx
tritonJS826: it is not db

mw-webapp/src/utils/auth/writeNewUserCredentialsToDataBase.ts
```diff
@@ -0,0 +1,30 @@
+import {getRedirectResult} from "firebase/auth";
+import {auth} from "src/firebase";
+import {UserService} from "src/service/UserService";
+
+const newUserBlankWays = {
+  ownWays: [""],
+  favoriteWays: [""],
+  mentoringWays: [""],
+};
```
tritonJS826, line 9: transfer

tritonJS826 (approved):

mw-webapp/src/utils/auth/writeNewUserCredentialsToDataBase.ts
SergioAJS: 1. Change name to writeNewUserCredentials
2. Transfer all User data in one block

PR #68, opened 2023-09-19
feat: add component Tooltip in the project

Arterixs:
1. Сreate Tooltip component
2. Add package ``clsx``


mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+} // In my opinion, this should be a global one that controls the naming of props
+
+interface TooltipProps {
+  children: ReactNode;
+  content: string;
+  styleComponent: ClassesTooltip;
+  position?: PositionTooltip;
+}
+
+export const Tooltip = ({children, content, styleComponent, position = PositionTooltip.TOP}: TooltipProps) => {
```
Ekaterina1994, line 22: Don't use destructure

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [visible, setVisible] = useState(false);
```
Ekaterina1994, line 23: if variable has boolean value better use naming with start is (isVisible)

Ekaterina1994 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+  TOP = "top",
+  LEFT = "left",
+  RIGHT = "right",
+  BOTTOM = "bottom"
+} // In my opinion, this should be a global one that controls the naming of props
+
+const enum ClassesTooltip {
+  MAIN = "main"
+} // In my opinion, this should be a global one that controls the naming of props
+
```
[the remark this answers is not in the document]
Arterixs: I didn’t find where to combine types in ``projectStructure.md``. I usually create a ``types`` folder in the root of the project, what should I do now?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface TooltipProps {
+  children: ReactNode;
+  content: string;
+  styleComponent: ClassesTooltip;
+  position?: PositionTooltip;
+}
+
+export const Tooltip = ({children, content, styleComponent, position = PositionTooltip.TOP}: TooltipProps) => {
+  const [visible, setVisible] = useState(false);
+  const classes = `${styles[styleComponent]} ${styles[position]}`; // I recomended use package 'clsx'
```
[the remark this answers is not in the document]
Arterixs: Should I install the package?

mw-webapp/src/component/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
tritonJS826: I think it's ok to put them in the current file, or nearby

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+} // In my opinion, this should be a global one that controls the naming of props
+
+const enum ClassesTooltip {
+  MAIN = "main"
+} // In my opinion, this should be a global one that controls the naming of props
+
+interface TooltipProps {
+  children: ReactNode;
+  content: string;
+  styleComponent: ClassesTooltip;
```
tritonJS826, line 18: nice prop to handle different types of tooltip! But currently we need just something simple, get rid of it)

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+  position?: PositionTooltip;
+}
+
+export const Tooltip = (props: TooltipProps) => {
+  const [isVisible, setVisible] = useState(false);
+  const positionTooltop = `${styles[props?.position ? props.position : PositionTooltip.TOP]}`;
+  const classes = `${styles[props.styleComponent]} ${styles[positionTooltop]}`; // I recomended use package 'clsx'
```
tritonJS826, line 25: 1. I recommend to use clsx too) 
2. put comments before commented lines, not after, it is standard practice

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface TooltipProps {
+  children: ReactNode;
+  content: string;
+  styleComponent: ClassesTooltip;
+  position?: PositionTooltip;
+}
+
+export const Tooltip = (props: TooltipProps) => {
+  const [isVisible, setVisible] = useState(false);
+  const positionTooltop = `${styles[props?.position ? props.position : PositionTooltip.TOP]}`;
```
tritonJS826, line 24: use `??` operator

tritonJS826, line 24: I dont see styles style[PositionTooltip.TOP] -> style['top'], what does this line do?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface TooltipProps {
+  children: ReactNode;
+  content: string;
+  styleComponent: ClassesTooltip;
+  position?: PositionTooltip;
+}
+
+export const Tooltip = (props: TooltipProps) => {
+  const [isVisible, setVisible] = useState(false);
```
tritonJS826, line 23: what about using plain css instead of react state? Is it possible?

mw-webapp/src/component/tooltip/Tooltip.tsx
tritonJS826, line 18: add some docs to your props please

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Ok, I install clsx package

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Ok, fixed

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: fixed

Arterixs: This line set position tooltip relative to cursor.
I add styles in scss file.

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Certainly, I fix it

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Ok, fix

mw-webapp/src/component/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
Arterixs: I moved the enum to an folder nearby to component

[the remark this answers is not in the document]
tritonJS826: Remember about encapsulation! Put relative things nearby) why do you need to separate folder for this enums?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /**
+  * Tooltip's content (text)
+  */
+  content: string;
+  /**
+  * Tooltip's class-style (selector css)
+  */
+  styleComponent: string;
+  /**
+  * Tooltip's position relative cursor (look enum PositionTooltip, optional, default - positon top)
```
tritonJS826, line 20: please, mention default value for optional props

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  */
+  children: ReactNode;
+  /**
+  * Tooltip's content (text)
+  */
+  content: string;
+  /**
+  * Tooltip's class-style (selector css)
+  */
+  styleComponent: string;
```
tritonJS826, line 18: let's use name className and appropriate comment

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface TooltipProps {
+  /**
+  * Tooltip's children (ReactNode)
+  */
+  children: ReactNode;
+  /**
+  * Tooltip's content (text)
+  */
+  content: string;
```
tritonJS826, line 14: looks on implementation it also could be ReactNode. Please rename content to text or change type to string|ReactNode

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /**
+  * Tooltip's class-style (selector css)
```
tritonJS826, line 16: why do you use synonyms for description in parenthesis? Choose the only one definition to your documentation

mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+.tooltip {
+  display: none;
+  position: absolute;
+  overflow: hidden;
+
+  &_top {
+    bottom: 130%;
```
tritonJS826, line 19: why do you count distance from bottom if it calls top? What will happened if your component height is 10vh?

mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,41 @@
...
+    transform: translateX(-50%);
```
tritonJS826, line 20: could you explain how this two lines works? why do you need transform and left at the same time?

mw-webapp/src/component/tooltip/positionTooltip/PositionTooltip.ts
```diff
@@ -0,0 +1,6 @@
+export const enum PositionTooltip {
+  TOP = "tooltip_top",
```
tritonJS826, line 2: 1.let's avoid BEM, let's use cameCase for className, it will be easier to work with them in tsx 
2. let 's use the same key and value for enums if possible

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/positionTooltip/PositionTooltip.ts
Arterixs: 1. ok
2. fixed

mw-webapp/src/component/tooltip/Tooltip.module.scss
Arterixs: ``left: 50%`` - centers content in the center ``.wrapper``
``transform: translateX(-50%)`` - makes the content center and wrapper center coaxial

mw-webapp/src/component/tooltip/Tooltip.module.scss
Arterixs: 1. I count distance from bootom if I calls top, to avoid writing negative values, for example: ``top: -80%``
2. If my component is given a height to ``vh``, then the display height of the component will depend on the height of the screen. In this case, when specifying ``%``, the height depends solely on the parent block

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: ok

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: ok, fixed

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: fixed

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: ok

mw-webapp/src/component/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
Arterixs: This enum is supposed to be imported into other parts of the code to indicate the component positioning to be applied. Where is the best place to put it? Create a file near to file the component? Or should I export it from this component file?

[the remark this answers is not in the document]
tritonJS826: I think it's ok to put it near to file with component

mw-webapp/src/component/tooltip/Tooltip.module.scss
tritonJS826: 1. what the problem with negative numbers?
2.  i mean how it will look? 30% above tall block means the tooltip will appeared somewhere in the sky, right)?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  */
+  children: ReactNode;
+  /**
+  * Tooltip's content (string | ReactNode)
+  */
+  content: string | ReactNode;
+  /**
+  * Tooltip's class-style (string)
+  */
+  className: string;
```
tritonJS826, line 18: what if I don't want to use custom styles for tooltip? Why is it required?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
+import {ReactNode} from "react";
+import clsx from "clsx";
+import {PositionTooltip} from "src/component/tooltip/positionTooltip/PositionTooltip";
+import styles from "src/component/tooltip/Tooltip.module.scss";
+
+interface TooltipProps {
+  /**
+  * Tooltip's children (ReactNode)
+  */
+  children: ReactNode;
```
tritonJS826, line 10: let's use generic React.FC to get rid of children prop in the interface

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /**
+  * Tooltip's content (string | ReactNode)
```
tritonJS826, line 12: why do you duplicate types in comments?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  */
+  content: string | ReactNode;
+  /**
+  * Tooltip's class-style (string)
+  */
+  className: string;
+  /**
+  * Tooltip's position relative cursor, optional (default - PositionTooltip.TOP)
```
tritonJS826, line 20: You have ts-doc features like {@link something} etc. Use it and will not have to read and syncronize all project docs after some variable renamings)

mw-webapp/src/component/tooltip/positionTooltip/PositionTooltip.ts
```diff
@@ -0,0 +1,6 @@
+export const enum PositionTooltip {
```
tritonJS826, line 1: 1. please add some documentation to all public members
2. PositionTooltip.TOP is even more transparent than PositionTooltip.TOOLTIP_TOP. Get rid of useless words

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.module.scss
Arterixs: 1. Ok, I agree
2. Right. The shell wraps around the child and has the boundaries of that children. The tooltip will be positioned 30% above the child’s border. If this is critical, we have a choice of other positions: left, right, bottom

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: what should you write then? Can you post a link that describes rules like documenting code?

mw-webapp/src/component/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
Arterixs: Done

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Done. Default className - tooltipDefault. 
I did lost class tooltipDefault empty, because I don't see markup.

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: In React 18, Children are no longer part of the FC type. It seems to me that this is already redundant typification
image

mw-webapp/src/component/tooltip/positionTooltip/PositionTooltip.ts
Arterixs: 1. Done
2. Done

mw-webapp/src/component/tooltip/Tooltip.tsx
Howl404: Just a short description

tritonJS826: ask @Ekaterina1994 in telegram to sent you a McConnel

mw-webapp/src/component/tooltip/Tooltip.tsx
tritonJS826: PositionTooltip.TOP is not a text it is a part of code. Link it

mw-webapp/src/component/tooltip/PositionTooltip.ts
```diff
@@ -0,0 +1,18 @@
...
+  /**
+  * Position Tooltip is left relative cursor
+  */
+  LEFT = "tooltipLeft",
+  /**
+  * Position Tooltip is right relative cursor
+  */
+  RIGHT = "tooltipRight",
+  /**
+  * Position Tooltip is bottom relative cursor
```
tritonJS826, line 15: 1. does your tooltip appears relatively to cursor? Or relatively to element under cursor?
2. PositionTooltip is a public member too
3. Why not to use same key and value, like TOP = 'top' or even TOP = 'TOP'

mw-webapp/src/component/tooltip/Tooltip.module.scss
tritonJS826: OK so if you will have two elements in a row one tall (text column 2000px) and one with small height (image in this text 50px), then tooltipTop will appeared for one of them 600px above (higher than header) and for  another 15px above. Is it your idea?

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
tritonJS826: ok what about generic `PropsWithChildren` in this case? I want to avoid mentioning children in props

Arterixs: Ok, done

mw-webapp/src/component/tooltip/Tooltip.module.scss
Arterixs: I got it. In this case, it is better to specify absolute values. Or for such a case, make your own different positioning. Then you need to make one default positioning and remove enam to set custom positioning depending on the case.

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: I got it. Done

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Done)

mw-webapp/src/component/tooltip/PositionTooltip.ts
Arterixs: 1. Done
2. I don't understand, what do you mean?
3. Done

tritonJS826: 2. All public members should be documented. You document fields in this interface but not interface itself.

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {PositionTooltip} from "src/component/tooltip/PositionTooltip";
+import styles from "src/component/tooltip/Tooltip.module.scss";
+
+interface TooltipProps {
+  /**
+  * Tooltip's content (text or JSX.Element)
+  */
+  content: string | ReactNode;
+  /**
+  * Tooltip's className, default ("default")
```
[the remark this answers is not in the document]
Arterixs: When there is a design layout, it will be possible to write default styles into this class

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+  */
+  className?: string;
+  /**
+  * Tooltip's position default {@link PositionTooltip.TOP}
+  */
+  position?: PositionTooltip;
+}
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = (props: PropsWithChildren<TooltipProps>) => {
+  const styleTooltip = styles[props.className ?? "default"];
```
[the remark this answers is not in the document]
Arterixs: I got it, fixed

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+  * Tooltip's className, default ("default")
+  */
+  className?: string;
+  /**
+  * Tooltip's position default {@link PositionTooltip.TOP}
+  */
+  position?: PositionTooltip;
+}
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = (props: PropsWithChildren<TooltipProps>) => {
```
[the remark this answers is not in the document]
Arterixs: I described the public method as I understood it, I didn’t find such rules, can you tell me what it’s called?

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+  /**
+  * Tooltip's content (text or JSX.Element)
+  */
+  content: string | ReactNode;
+  /**
+  * Tooltip's className, default ("default")
+  */
+  className?: string;
+  /**
+  * Tooltip's position default {@link PositionTooltip.TOP}
```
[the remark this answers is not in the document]
Arterixs: Done

mw-webapp/src/component/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
Ekaterina1994: Let's keep it to task #86

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * Describe props component Tooltip
+ */
+interface TooltipProps {
+  /**
+  * Tooltip's content
+  */
+  content: string | ReactNode;
+  /**
+  * Tooltip's className
+  * default ("tooltip")
```
tritonJS826, line 16:
```suggestion
```

tritonJS826, line 16: it will not be tooltip because you are using module.scss. Lets incapsulate this data. I added github suggestion

tritonJS826 (approved):

mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.tooltip {
+  display: none;
+  position: absolute;
+  overflow: hidden;
+}
+
+.top {
+  transform: translateY(-102%) translateX(-50%);
```
tritonJS826, line 20: now your surrogate "margin" will be 2%  (translate), it depends on content and will be look not consistent, better to use absolute values in case like this, agreed?

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: I get it)

mw-webapp/src/component/tooltip/Tooltip.module.scss
Arterixs: I agree, correct

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
+import {PropsWithChildren, ReactNode} from "react";
+import clsx from "clsx";
+import {PositionTooltip} from "src/component/tooltip/PositionTooltip";
+import styles from "src/component/tooltip/Tooltip.module.scss";
+
+/**
+ * Describe props component Tooltip
```
tritonJS826, line 7:
```suggestion
 * Tooltip props
```

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  className?: string;
+  /**
+  * Tooltip's position
+  * default: {@link PositionTooltip.TOP}
+  */
+  position?: PositionTooltip;
+}
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = (props: PropsWithChildren<TooltipProps>) => {
+  const classes = clsx(styles.tooltip, props.className && styles[props.className], styles[props.position ?? PositionTooltip.TOP]);
```
tritonJS826, line 26: What does second parameter do? Did you check it?

tritonJS826 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: The second condition implies the presence of a classname prop, and if it is a formalized style conveyed in this prop, if not, then it simply skips it

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  */
+  className?: string;
+  /**
+  * Tooltip's position
+  * default: {@link PositionTooltip.TOP}
+  */
+  position?: PositionTooltip;
+}
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = (props: PropsWithChildren<TooltipProps>) => {
```
Ekaterina1994, line 25: Where is ts-doc for public member?

mw-webapp/src/component/tooltip/Tooltip.tsx
Ekaterina1994: If you will not get prop className then styles[props.className]  will not be available

Ekaterina1994 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: Right! Notice, ``classname`` is optional type, custom style. Default style is `tooltip`

mw-webapp/src/component/tooltip/Tooltip.tsx
Arterixs: I don’t understand, when entering an example in the code, how it should look like, if I open any component in the project, it won’t be there, should I @public tag the function? Or are you talking about a property in the interface?

tritonJS826 (approved):

Ekaterina1994 (approved):

PR #72, opened 2023-09-19
feat: #58 add storybook to project 

Howl404:


mw-webapp/vite.config.ts
```diff
@@ -30,7 +30,7 @@ export default defineConfig(() => {
 
   return {
     build: {outDir: "build"},
-    plugins: [react(), eslint(), viteTsconfigPaths()],
+    plugins: [react(), eslint({exclude: ["/virtual:/**"]}), viteTsconfigPaths()],
```
tritonJS826, line 33: why do we need `{exclude: ["/virtual:/**"]}`? Could you add a comment please?

mw-webapp/src/stories/button/Button.stories.ts
```diff
@@ -0,0 +1,17 @@
+import {Button} from "./Button";
```
tritonJS826, line 1: is it possible to use absolute path here?

mw-webapp/src/stories/Configure.mdx
tritonJS826, line 1: We are not going to create separate stories without components in nearest future, so could we get rid of directory `stories`, looks like it will be empty?

tritonJS826 (changes requested):

mw-webapp/src/stories/Configure.mdx
```diff
@@ -0,0 +1,364 @@
+import { Meta } from "@storybook/blocks";
```
tritonJS826, line 1: We are not going to create separate stories without components in nearest future, so could we get rid of directory stories, looks like it will be empty?

mw-webapp/vite.config.ts
Howl404: Because of this error - 
image

Found only this fix - https://github.com/storybookjs/builder-vite/issues/535#issuecomment-1507352550

mw-webapp/src/stories/Configure.mdx
Howl404: I was told to move Button into stories, will putt it back now

mw-webapp/src/stories/button/Button.stories.ts
Howl404: Done

mw-webapp/src/stories/Configure.mdx
Howl404: Done

mw-webapp/vite.config.ts
tritonJS826: Ok, could you add a comment to the code to document it? With this link if possible. It will save time next for future developers.

mw-webapp/src/stories/Configure.mdx
tritonJS826: Who and where told you to put it to the stories folder?

tritonJS826: Yeah, sorry for the misunderstanding, our fault. Thanks for notifying

mw-webapp/vite.config.ts
Howl404: Done

Ekaterina1994 (approved):

PR #77, opened 2023-09-20
refactor: #39 move out network related methods from the utils folder

SergioAJS:


mw-webapp/src/service/auth/logIn.ts
```diff
@@ -1,7 +1,7 @@
 import {signInWithRedirect} from "firebase/auth";
 import {auth, provider} from "src/firebase";
 
-export const handleLogIn = async () => {
+export const logIn = async () => {
```
Howl404, line 4: What is the reason to change function names? They are still handlers in my opinion. handleLogIn -> signInWithRedirect, looks better than logIn -> signInWithRedirect

mw-webapp/src/component/header/Header.tsx
```diff
@@ -28,7 +28,7 @@ export const Header = () => {
       </h1>
       <div className={styles.block_button}>
```
Ekaterina1994, line 29: rename classname . we don't use _ for classes. Use camelCase

mw-webapp/src/service/auth/logIn.ts
Ekaterina1994: handle is too general word. So we try don't use it for specific functions. 
Login is well

mw-webapp/src/service/auth/logIn.ts
```diff
@@ -1,7 +1,7 @@
...
   try {
     await signInWithRedirect(auth, provider);
```
Ekaterina1994, line 6: is it possible to rename it to logInWithRedirect?

Ekaterina1994 (changes requested):

mw-webapp/src/service/auth/logIn.ts
tritonJS826: the main question here does it (using word handle) make code more readable. If no - we should not use it

mw-webapp/src/pages/authPage/AuthPage.tsx
```diff
@@ -16,7 +16,7 @@ export const AuthPage = () => {
   const [user, setUser] = useState<User | null>(null);
 
   useEffect(() => {
-    handleUserAuthState(setUser);
+    userAuthState(setUser);
     writeNewUserCredentials();
```
tritonJS826, line 20: writeNewUserCredentials calls on mounting AuthPage and it set data to user:
`
ownWays: [""],
favoriteWays: [""],
mentoringWays: [""],
` 
Looks like it will reset all my favorite ways every time I'm logged in, it will be so sad for users)
What do you think?

tritonJS826 (changes requested):

mw-webapp/src/service/auth/logIn.ts
SergioAJS: No, this is an internal firebase method. That's why I named function and button value as signIn for the first time.

mw-webapp/src/component/header/Header.tsx
SergioAJS: Done

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Yes, you are right.
I will set only the name, email, and uid for now. Later we can write logic to check if the user uid is already in Firebase and then keep his ways.
Maybe don't get a user name from Google also? If the user can change his name on our site then when the user login again, the name will also reset. What do you think?

tritonJS826: Yeah, it looks better! Also we could check firebase docs, maybe they provide appropriate method.

mw-webapp/src/service/auth/logIn.ts
Howl404: I think try\catch block can be considered handling?

mw-webapp/src/component/header/Header.tsx
```diff
@@ -18,17 +18,17 @@ export const Header = () => {
   const [user, setUser] = useState<User | null>(null);
 
   useEffect(() => {
-    handleUserAuthState(setUser);
+    userAuthState(setUser);
```
tritonJS826, line 21: functions should contain a verb because it should do something

mw-webapp/src/pages/authPage/AuthPage.tsx
tritonJS826, line 20: we should call this function only if user dont exist

mw-webapp/src/service/auth/logIn.ts
tritonJS826: let's skip check services to errors, we will improve it latere

mw-webapp/src/component/header/Header.tsx
SergioAJS: Add verb handle to the name fo the function

mw-webapp/src/pages/authPage/AuthPage.tsx
SergioAJS: Move adding a new user to DB logic and checking if a user already exists to  handleUserAuthState function

SergioAJS: Done

Ekaterina1994 (approved):

Howl404 (approved):

mw-webapp/src/service/auth/logIn.ts
```diff
@@ -0,0 +1,6 @@
+import {signInWithRedirect} from "firebase/auth";
+import {auth, provider} from "src/firebase";
+
+export const logIn = async () => {
```
tritonJS826, line 4: !

tritonJS826 (approved):

tritonJS826 (approved):

PR #79, opened 2023-09-20
feat: #56 added 404 page

chelovek-zeml:


mw-webapp/src/pages/Page404/Page404.tsx
```diff
@@ -0,0 +1,11 @@
+import styles from "src/pages/Page404/Page404.module.scss";
+
+const ERROR_404 = "404 NOT FOUND";
```
Howl404, line 3: In this case, where string is only used in one place and is unlikely to be reused, i don't see a reason to make it a variable, little complexity without benefit

mw-webapp/src/styles/_variables.scss
```diff
@@ -12,4 +12,11 @@ $link-active: rgb(255, 149, 0);
 $header-background: rgb(85, 122, 149);
 
 // Title
-$title-padding: 10px;
\ No newline at end of file
+$title-padding: 10px;
+
+// font size
```
Howl404, line 17: Start comment from uppercase

mw-webapp/src/pages/Page404/Page404.module.scss
```diff
@@ -0,0 +1,10 @@
+@import "src/styles/variables.scss";
+
+.container {
+  display: flex;
+  width: 100%;
+  height: 70vh;
```
Ekaterina1994, line 6: why height of page's container is 70vh?

Ekaterina1994 (changes requested):

mw-webapp/src/pages/Page404/Page404.tsx
Howl404: I think using a variable here is breaking KISS.

1. String appears only once
2. String will never change
3. Meaning of the string is obvious from the context

By this logic we should make every simple html text into variables

mw-webapp/src/styles/_variables.scss
chelovek-zeml: Fixed.

mw-webapp/src/styles/_variables.scss
```diff
@@ -12,4 +12,11 @@ $link-active: rgb(255, 149, 0);
...
+$font-s-small: 14px;
+$font-s-regular: 16px; // -> base size
+$font-s-medium: 18px; 
+$font-s-big: 20px;
+$font-s-large: 48px;
```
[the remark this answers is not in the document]
chelovek-zeml: s - means size(font-size).

mw-webapp/src/pages/Page404/Page404.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const Page404 = () => {
```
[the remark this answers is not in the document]
chelovek-zeml: You can't start a function name with a number.

mw-webapp/src/pages/Page404/Page404.module.scss
chelovek-zeml: If you take the entire height of the screen, scrolling appears.

mw-webapp/src/pages/Page404/Page404.tsx
chelovek-zeml: It's useful because this text is a magic value. All magic values should be move out in consts.

mw-webapp/src/pages/Page404/Page404.tsx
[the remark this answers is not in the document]
chelovek-zeml: can be named constant: TEXT_ERROR?

mw-webapp/src/pages/Page404/Page404.tsx
Howl404: "A "magic value" is a string or number used in a program that is essential to its proper function but provides no context or explanation for why it is what it is."

Only definition that i found, and by it - you cant really call this string a magic value, context is sufficient, will be happy to see other definition

mw-webapp/src/pages/Page404/Page404.module.scss
tritonJS826: page should have dynamic height depends on height of content

mw-webapp/src/pages/Page404/Page404.tsx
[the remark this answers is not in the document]
tritonJS826: 1. On my own opinion numbers at the end looks ok
2. could you add a link to main page to get ability to move on for users)

mw-webapp/src/pages/Page404/Page404.tsx
```diff
@@ -0,0 +1,11 @@
+import styles from "src/pages/Page404/Page404.module.scss";
+
+const TEXT_ERROR = "404 NOT FOUND";
+
+export const Page404 = () => {
+  return (
+    <div className={styles.container}>
```
tritonJS826, line 7: I propose to use as different naming if possible. For example almost each component could contain container and it will be a little hard to find appropriate one when you are trying to debug etc. Much more easier to find  something more specific like errorPageContainer. It is just matter of code style, let's try this one

mw-webapp/src/styles/_variables.scss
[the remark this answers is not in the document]
tritonJS826: 1. is it possible to use cameCase for variables?
2. please put comment before lines
3. Do we really need 5 sizes? get rid of unused

mw-webapp/src/pages/Page404/Page404.tsx
tritonJS826: @Howl404 agreed with you, but it is just a convention for the current project to move all these kinds of values to variables. In my opinion, it is simpler to put all variables in constants than teach everybody to differentiate without debates about what will be used somewhere else and what is the real magic number. BTW ,  you provide good 3 points above, I like them :) You could provide clear convention for what should be placed in separate variable and what should not and we can discuss  it later

mw-webapp/src/styles/_variables.scss
tritonJS826: Should linter fix it automatically? Please check configuration

mw-webapp/src/pages/Page404/Page404.tsx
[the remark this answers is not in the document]
tritonJS826: dont see changes related to point 2

tritonJS826 (changes requested):

tritonJS826:
I cant approve until you merge conflicts

tritonJS826:
Don't see any changes related to the last review

mw-webapp/src/pages/Page404/Page404.tsx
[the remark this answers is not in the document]
tritonJS826: Still no changes

tritonJS826 (changes requested):

chelovek-zeml:
I wrote above that we have a common header and there is a link to the main page there.

mw-webapp/src/pages/Page404/Page404.tsx
```diff
@@ -0,0 +1,11 @@
+import styles from "src/pages/page404/Page404.module.scss";
+
+const ERROR_404 = "404 NOT FOUND";
+
+export const Page404 = () => {
+  return (
+    <div className={styles.errorPageContainer}>
```
tritonJS826, line 7: Semantically each page should contain a heading, please add it before merging

mw-webapp/src/styles/_variables.scss
```diff
@@ -13,3 +13,6 @@ $header-background: rgb(85, 122, 149);
 
 // Title
 $title-padding: 10px;
+
+// Font size
+$font-size-large: 48px;
```
tritonJS826, line 18: What about font size for title? Use mixin for title before merge

tritonJS826 (approved):

chelovek-zeml:
Done

mw-webapp/src/styles/_variables.scss
Howl404: 1

mw-webapp/src/pages/Page404/Page404.tsx
```diff
@@ -0,0 +1,16 @@
+import {Heading} from "@radix-ui/themes";
+import {HeadingLevel} from "src/component/title/Title";
+import styles from "src/pages/page404/Page404.module.scss";
```
Ekaterina1994, line 3: We have component title for heading/ No need create new component

Ekaterina1994 (changes requested):

Ekaterina1994 (approved):

PR #102, opened 2023-09-28
feat: #92 add notice component without styles and storyBook

anterebol:
The StoryBook is in process...

### DIFF
--- a/mw-webapp/src/component/notice/Notice.tsx
+++ b/mw-webapp/src/component/notice/Notice.tsx
@@ -1,0 +1,71 @@
+import {useEffect} from "react";
+import {Button} from "../button/Button";
+import {Title} from "../title/Title";
+import {HeadingLevel} from "../title/Title";
+import styles from "./Notice.module.scss";
+
+const MILLISECONDS = 1000;
+
+export interface NoticeProps {
+  /**
+   * ID for notification, it will be help closing notice by button (number, which we will be use like string)
+   */
+  id: string;
+  /**
+  * Title for notice (text)
+  */
+  title: string;
+  /**
+   * Description for notice (text)
+   */
+  description: string;
+  /**
+   * Type for notice, like 'Error', 'Ok', 'Warning' (text)
+   */
+  type: string;
+  /**
+   * Timer for automaticle closing (seconds)
+   */
+  timer: number;
+  /**
+   * Callback which close the notification
+   */
+  closeNotice: (id: string) => void;
+}
+
+export const Notice = (props: NoticeProps) => {
+
+  const {id, title, description, type, timer, closeNotice} = props;
+
+  const millisecondsTimer = timer * MILLISECONDS;
+
+  useEffect(() => {
+    setTimeout(() => closeNotice(id), millisecondsTimer);
+  }, []);
+
+  return (
+    <article
+      className={[styles.notice, styles[type]].join(" ")}
+      id={id}
+    >
+      <div className={styles.crose_box}>
+        <img
+          src="./"
+          alt="close_img"
+        />
+      </div>
+      <Title
+        level={HeadingLevel.h3}
+        className={styles.title}
+        text={title}
+      />
+      <p className={[styles.description].join(" ")}>
+        {description}
+      </p>
+      <Button
+        value={"Ok"}
+        onClick={() => closeNotice(id)}
+      />
+    </article>
+  );
+};
--- a/mw-webapp/src/component/notice/Notice.tsx
+++ b/mw-webapp/src/component/notice/Notice.tsx
@@ -8,7 +8,7 @@ const MILLISECONDS = 1000;
 
 export interface NoticeProps {
   /**
-   * ID for notification, it will be help closing notice by button (number, which we will be use like string)
+   * ID for notification, it will be help closing notice by button
    */
   id: string;
   /**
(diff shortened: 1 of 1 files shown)


mw-webapp/src/component/notice/Notice.module.scss
Ekaterina1994, line 1: If you don't use styles for component then delete this file

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
+import {useEffect} from "react";
+import {Button} from "../button/Button";
+import {Title} from "../title/Title";
+import {HeadingLevel} from "../title/Title";
+import styles from "./Notice.module.scss";
```
Ekaterina1994, line 5: WE use absolute paths everywhere

Ekaterina1994, line 5: We use absolute paths everywhere

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const MILLISECONDS = 1000;
+
+export interface NoticeProps {
```
Ekaterina1994, line 9: Add js-doc to each export public members

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
+import {useEffect} from "react";
+import {Button} from "../button/Button";
+import {Title} from "../title/Title";
+import {HeadingLevel} from "../title/Title";
+import styles from "./Notice.module.scss";
+
+const MILLISECONDS = 1000;
```
Ekaterina1994, line 7: We no need this const in components. Id you will need it you can get it by props from logic layer. Am I wrong?

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+   */
+  description: string;
+  /**
+   * Type for notice, like 'Error', 'Ok', 'Warning' (text)
+   */
+  type: string;
+  /**
+   * Timer for automaticle closing (seconds)
+   */
+  timer: number;
```
Ekaterina1994, line 29: Do we always need this property? Maybe better make it optional. For example I need to read notification few times? then better don't close it automatically

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Callback which close the notification
+   */
+  closeNotice: (id: string) => void;
+}
+
+export const Notice = ({id, title, description, type, timer, closeNotice}: NoticeProps) => {
```
Ekaterina1994, line 36: 1/ add js-doc to public member
2/ don't use destructure in props. Image that you have 100 props/ It's will hard to write everything

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+        level={HeadingLevel.h3}
+        className={styles.title}
+        text={title}
+      />
+      <p className={[styles.description].join(" ")}>
+        {description}
+      </p>
+      <Button
+        value={"Ok"}
+        onClick={() => closeNotice(id)}
```
Ekaterina1994, line 58: Can't we write onClick={closeNotice(id)} ?

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Title
+        level={HeadingLevel.h3}
+        className={styles.title}
+        text={title}
+      />
+      <p className={[styles.description].join(" ")}>
+        {description}
+      </p>
+      <Button
+        value={"Ok"}
```
Ekaterina1994, line 57: If value === "OK" everywhere then make it const if value can be difference - need get it from props

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <article
+      className={[styles.notice, styles[type]].join(" ")}
+      id={id}
+    >
+      <Title
+        level={HeadingLevel.h3}
+        className={styles.title}
+        text={title}
+      />
+      <p className={[styles.description].join(" ")}>
```
Ekaterina1994, line 53: Can you please explain what this class means? Why you need join?

mw-webapp/src/component/notice/Notice.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const Notice = ({id, title, description, type, timer, closeNotice}: NoticeProps) => {
+
+  const millisecondsTimer = timer * MILLISECONDS;
+
+  useEffect(() => {
+    setTimeout(() => closeNotice(id), millisecondsTimer);
+  }, []);
+  return (
+    <article
```
Ekaterina1994, line 44: Why tag article?

Ekaterina1994 (changes requested):

mw-webapp/src/component/notice/Notice.tsx
anterebol: Ok, will change it.

mw-webapp/src/component/notice/Notice.tsx
anterebol: Sure. I'm agree with you.

mw-webapp/src/component/notice/Notice.tsx
anterebol: I thougt to write semantic tag. Do you prepare usual "div"

mw-webapp/src/component/notice/Notice.tsx
anterebol: Sorry, it's my mistake, forgot to delete.

mw-webapp/src/component/notice/Notice.tsx
anterebol: We receive timer (seconds) inside the props, but setTimeout use miliseconds inside. We just convert it.

mw-webapp/src/component/notice/Notice.tsx
anterebol: I would suggest using the cross button in the angle, but for it I think we need to create another component.

mw-webapp/src/component/notice/Notice.tsx
Ekaterina1994: No, no div. Just ask is it the best semabtic tag for notifications

PR #106, opened 2023-09-28
feat: #82 add image component

Pivodur:
### Overview
added an image component

### Changes description
- image component was implemented 
- storybook for image was implmented


mw-webapp/src/component/image/Image.stories.ts
```diff
@@ -0,0 +1,14 @@
...
+  tile: "Image",
+  component: Image,
+  parameters: {layout: "centered"},
+  tags: ["autodocs"],
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {args: {src: "https://picsum.photos/200/300", alt: "Image"}};
```
tritonJS826, line 14: wow) can you guarantee that this image will be exist tomorrow? What happened with preview in story book if image disappeared? it's gonna surprise the developers a little bit.

tritonJS826: why alt - image? does it semantically correct?

mw-webapp/src/component/image/Image.tsx
```diff
@@ -0,0 +1,30 @@
...
+   * Image alt text
+   */
+  alt: string;
+  /**
+   * Additional custom class name for the component
+   */
+  className?: string;
+}
+
+export const Image: React.FC<ImageProps> = (props: ImageProps) => {
```
tritonJS826, line 19: Please read docs again, all public members should be documented (all exported functions, variables etc.
)

tritonJS826 (changes requested):
Good luck)

mw-webapp/src/component/image/Image.stories.ts
Pivodur: 1. I can't guarantee that the source of the image will still exist. But what can I use as a preview instead?
2. Do you want me to write more descriptive text like "Default storybook picture"?

Ekaterina1994: 1. You can download any picture and use it for story
2. Yes need to add description but it depends on which picture.

mw-webapp/src/component/image/Image.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import styles from "src/component/image/Image.module.scss";
+
+interface ImageProps {
```
Ekaterina1994, line 4: Add js-doc to interface too

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.stories.ts
Pivodur: Should I place an image in the same folder of the component, or should I create an 'assets' folder?

Ekaterina1994: Create folder for images

Issue #65, opened 2023-09-17
Use flag instead of aray[0]


Ekaterina1994:
mw-webapp/src/component/table/columns.tsx
String 31: div className={arrayItem[FIRST_INDEX] === "✓" ? styles.completed : styles.notCompleted}

too raugh check by first char in string, better to use flag

Use flag (property true/false) instead of finding symbol for mentorComment

[] Update models in draw.io
[] Update firebase collection
[] Update business models
[] Check value of property to style elements

PR #107, opened 2023-09-30
feat: #65 Add comments

Howl404:


mw-webapp/src/App.tsx
```diff
@@ -1,6 +1,9 @@
 import {Header} from "src/component/header/Header";
 import {Router} from "src/router/Router";
 
+/**
+ * App
+ */
 function App() {
   return (
```
Ekaterina1994, line 8: Empty line before return

Ekaterina1994 (changes requested):

tritonJS826 (approved):

mw-webapp/src/App.tsx
Howl404: { "blankLine": "always", "prev": "*", "next": "return" }

I added this rule, but seems like ESLint do not count whole function as something and only add blankline between something inside it

Howl404: Added

SergioAJS (approved):

Ekaterina1994 (approved):

Issue #132, opened 2023-10-13
Add cypress to project


Ekaterina1994:
Cypress components testing

Ekaterina1994:
Done

PR #134, opened 2023-10-14
feat: #132 add cypress in project

Ekaterina1994:


mw-webapp/src/component/button/Button.cy.tsx
```diff
@@ -0,0 +1,36 @@
+import {Button} from "src/component/button/Button";
+
+const BUTTON_VALUE = "Click button";
+
+describe("<Button />", () => {
```
tritonJS826, line 5: Let's skip test which check possibility to render

mw-webapp/src/component/button/Button.cy.tsx
```diff
@@ -0,0 +1,36 @@
...
+  it("renders", () => {
+    const STUB_FUNCTION = cy.stub();
+    cy.mount(
+      <Button
+        value={BUTTON_VALUE}
+        onClick={STUB_FUNCTION}
+      />);
+  });
+
+  it("uses custom text for the button", () => {
```
tritonJS826, line 15: What does mean custom? Do you have default? Describe what do you test like : something should happened in case if something... Etc.

tritonJS826 (approved):

tritonJS826 (commented):
Add appropriate description in describe too

SergioAJS (approved):

mw-webapp/src/component/button/Button.cy.tsx
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it("clicking", () => {
+    const STUB_FUNCTION = cy.stub();
+    cy.mount(
+      <Button
+        value={BUTTON_VALUE}
+        onClick={STUB_FUNCTION}
+      />);
+    cy.get("button").click().then(() => {
```
SergioAJS, line 32: Maybe better to use data-* attributes for selecting elements?
https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

Ekaterina1994: Changed. Thanks

mw-webapp/src/component/button/Button.cy.tsx
Ekaterina1994: Rewrote

mw-webapp/src/component/button/Button.cy.tsx
Ekaterina1994: Deleted

SergioAJS (approved):

tritonJS826 (approved):

PR #144, opened 2023-10-18
refactor: #143 Update models according drawio

Ekaterina1994:


Howl404 (approved):

mw-webapp/src/dataAccessLogic/getUsersPreview.ts
```diff
@@ -14,7 +14,8 @@ export const getUsersPreview = async (): Promise<UserPreview[]> => {
 
   const firstUser = usersDTO[0];
 
-  const ownWays = firstUser.ownWays.map((ownWayUuid) => {
+  // TODO: this file will be deleted after merge PR #95
```
tritonJS826, line 17: So what to do?

mw-webapp/src/logic/reportsTable/renderCellValue/renderCellItem.tsx
```diff
@@ -8,7 +8,7 @@ import {MentorComment} from "src/model/businessModel/MentorComment";
 import {PlanForNextPeriod} from "src/model/businessModel/PlanForNextPeriod";
 
 /**
- * Properties of {@link DayReport} that have type string[] for dynamic update
+ * Properties of DayReport model that have type string[] for dynamic update
```
tritonJS826, line 11: Why not to use @link

mw-webapp/src/model/businessModel/DayReport.ts
```diff
@@ -64,7 +64,7 @@ export class DayReport {
       new CurrentProblem(currentProblemItem));
     this.studentComments = dayReportData.studentComments;
     this.learnedForToday = dayReportData.learnedForToday;
-    this.mentorComments = dayReportData.mentorComments.map((mentorComment) =>
+    this.mentorComments = dayReportData.mentorComments?.map((mentorComment) =>
```
tritonJS826, line 67: How does it possible, why null?

tritonJS826 (approved):

mw-webapp/src/dataAccessLogic/getUsersPreview.ts
Ekaterina1994: Delete

mw-webapp/src/logic/reportsTable/renderCellValue/renderCellItem.tsx
Ekaterina1994: Bacuse we can links only to those models that was imported in file. Otherwise this link doesn't work

Issue #142, opened 2023-10-16
Create all ways table


Ekaterina1994:
on AllWaysPage

Ekaterina1994:
Done

PR #145, opened 2023-10-18
feat: #142 Create AllWays table

Ekaterina1994:


mw-webapp/src/dataAccessLogic/GoalPreviewDAL.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Provides methods to interact with the GoalPreview model
+ */
+export class GoalPreviewDAL {
+
+  /**
+   * Get GoalsPreview
+   */
+  public static async getGoalsPreview(): Promise<GoalPreview[]> {
+    const goalsDTO = await GoalService.getGoalsDTO();
+    const goalsPreview = goalsDTO.map((goalDTO) => (goalDTOToGoalPreviewConverter(goalDTO)));
```
tritonJS826, line 15: I think this syntax more readable, agreed? check everywherye
```suggestion
    const goalsPreview = goalsDTO.map(goalDTOToGoalPreviewConverter);
```

tritonJS826 (approved):
Cool, thanks!)

mw-webapp/src/dataAccessLogic/GoalPreviewDAL.ts
Ekaterina1994: Changed everywhere

mw-webapp/src/dataAccessLogic/DTOToPreviewConverter/userDTOToUserPreviewConverter.ts
```diff
@@ -1,15 +1,15 @@
 import {UserPreview} from "src/model/businessModelPreview/UserPreview";
-import {WayPreview} from "src/model/businessModelPreview/WayPreview";
+// Import {WayPreview} from "src/model/businessModelPreview/WayPreview";
```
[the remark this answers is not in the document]
Ekaterina1994: Deleted

SergioAJS (approved):

mw-webapp/src/logic/allWaysPage/allWaysTable/AllWaysTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {getCoreRowModel, useReactTable} from "@tanstack/react-table";
+import {Table} from "src/component/table/Table";
+import {WayPreviewDAL} from "src/dataAccessLogic/WayPreviewDAL";
+import {columns} from "src/logic/allWaysPage/allWaysTable/columns";
+import {WayPreview} from "src/model/businessModelPreview/WayPreview";
+
+/**
+ * Render table of all ways preview
+ */
+export const AllWaysTable = () => {
```
Howl404, line 11: Maybe make it reusable and instead of getting ways inside, pass them as param? Then we can use it in OwnWays, Favourite ways and etc. If not we will have duplicate code

Howl404: TLDR: Make it WaysTable with WayPreview[] as param

Ekaterina1994: 1. Sorry but we can't reusable any table with data because for each table will be different type of columns. We have reusable component Table. Doesn't it enough?
Maybe I didn't understood something?

2. About renaming. We have page AllWays so table should also be named AllWaysTable. Am I wrong?

Howl404: Colums can be same for AllWays, Owned Ways and etc, why not?

Howl404: What is the difference if we need to display WayPreview in all cases

Ekaterina1994: Yes we use one model nut the column can be difference/ I don't know. I didn't thought about tables in common yet. Created discuss in general)

Howl404 (approved):

PR #152, opened 2023-10-20
feat: #48 add table with mentoring ways to user page

Howl404:


mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
```diff
@@ -79,8 +79,19 @@ export class WayPreviewDAL {
   /**
    * Get User own ways preview
    */
-  public static async getOwnWaysPreview(uuid: string): Promise<WayPreview[]> {
-    const waysDTO = await WayService.getOwnWaysDTO(uuid);
+  public static async getUserWaysPreview(uuid: string, type: "OwnWays" | "MentoringWays"): Promise<WayPreview[]> {
+
```
Ekaterina1994, line 83: Add to comment that you get not only own ways but also mentoring ways

mw-webapp/src/logic/userPage/UserPage.tsx
```diff
@@ -17,11 +18,22 @@ export const UserPage = () => {
...
+              level={HeadingLevel.h3}
+            />
+            <OwnWaysTable uuid={uuid} />
+            <Title
+              text="Mentoring ways"
+              level={HeadingLevel.h3}
+            />
+            <MentoringWaysTable uuid={uuid} />
+          </>
+        )
```
Ekaterina1994, line 35: I think better move this logic to separate function

mw-webapp/src/logic/waysTable/MentoringWaysTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+   */
+  uuid: string;
+}
+
+/**
+ * Render table of mentoring ways preview
+ */
+export const MentoringWaysTable = (props: MentoringWaysTableProps) => {
+  const [ownWays, setOwnWays] = useState<WayPreview[]>([]);
+
```
Ekaterina1994, line 23: If you load mentoringWays why state calls ownWays

Ekaterina1994 (changes requested):

mw-webapp/src/logic/userPage/UserPage.tsx
Howl404: I think it's ok right now, another function is kinda unnecessary in my opinion

mw-webapp/src/logic/waysTable/MentoringWaysTable.tsx
Howl404: Already renamed

mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
Howl404: Done

SergioAJS (approved):

mw-webapp/src/logic/userPage/UserPage.tsx
tritonJS826: Personally I don't like ternarian operator is such cases. Also I think that we should not redirect from jsx, better to redirect on the next line after reading  params, it is about encapsulation, let's keep cohesive data and logic nearby

tritonJS826 (approved):

Ekaterina1994 (approved):

PR #155, opened 2023-10-22
feat: #55 add table with favourite ways to user page

Howl404:


mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
```diff
@@ -79,17 +79,20 @@ export class WayPreviewDAL {
   /**
    * Get User ways preview based of provided type
    */
-  public static async getUserWaysPreview(uuid: string, type: "OwnWays" | "MentoringWays"): Promise<WayPreview[]> {
+  public static async getUserWaysPreview(uuid: string, type: "Own" | "Mentoring" | "Favourite"): Promise<WayPreview[]> {
```
SergioAJS, line 82: "Favorite" is the right word. Please check the correct spelling of the word in all files

SergioAJS (changes requested):

mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
Howl404: [Favorite and Favourite are both correct](https://www.merriam-webster.com/grammar/favorite-or-favourite-usage)

SergioAJS (approved):

mw-webapp/src/service/WayService.ts
```diff
@@ -47,12 +48,25 @@ export class WayService {
...
     return mentoringWays;
   }
 
+  /**
+   * Get WaysDTO of user favourite ways
+   */
+  public static async getFavouriteWaysDTO(uuid: string): Promise<WayDTO[]> {
+    const userDTO = await UserService.getUserDTO(uuid);
+    const waysRef = collection(db, PATH_TO_WAYS_COLLECTION);
+    const favouriteWaysQuery = query(waysRef, where("uuid", "in", userDTO.favoriteWayUuids));
```
SergioAJS, line 65: On line 65, there are two spellings of the word "favorite". let's choose one variant

Howl404: Changed to favorite everywhere

tritonJS826 (approved):

mw-webapp/src/logic/waysTable/FavoriteWaysTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Favorite ways table props
+ */
+interface FavoriteWaysTableProps {
+
+  /**
+   * User Uuid
+   */
+  uuid: string;
+}
```
Ekaterina1994, line 16: Each table has the same interface with only one prop uuid/ What do you think about create one common props for all waysTables? Is it good or bad idea?

Ekaterina1994, line 16: Each table has the same interface with only one prop uuid/ What do you think about create one common props for all waysTables? Is it good or bad idea?

Ekaterina1994 (changes requested):

mw-webapp/src/logic/waysTable/FavoriteWaysTable.tsx
tritonJS826: if it is common you can even create a generic like WithChildrenProp, but for your purposes (for adding uuid)

Howl404: Maybe I'm not understanding something

Interface PropsWithUuid {
  uuid: string
}

Ekaterina1994: If speak about generics then maybe you can use :
type PropsWithUuid<T> = {
  uuid: T;
};

But with string it's more clear/ Don't know

Ekaterina1994 (changes requested):

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #41, opened 2023-09-13
Create styles architecture (if we need, choose design for the project) FOR UX/UI


Ekaterina1994:
- [ ] Choose 
- [ ] Create docs (conventions)
- [ ] Fix styles (create folders  + change styles in compoments , pages, etc)

Ekaterina1994:
Leave the task until the UX/UI designer arrives

PR #169, opened 2023-10-28
refactor: #159 change styles for header

MaratReactJs:


mw-webapp/src/styles/_variables.scss
```diff
@@ -10,6 +10,7 @@ $link-active: rgb(255, 149, 0);
 
 // Header
 $header-background: rgb(85, 122, 149);
+$header-padding: 0 1em;
```
Howl404, line 13: Why padding? It's looks uncentered

Howl404 (changes requested):

mw-webapp/src/styles/_variables.scss
MaratReactJs: and how it should be?

Ekaterina1994, line 13: What do zou think about using rem instead of em?

Ekaterina1994, line 13: If this padding will be used only in header then need move this var to header.scss. Better skip all styles changes on task #41

Ekaterina1994 (changes requested):

Ekaterina1994 (approved):

mw-webapp/src/styles/_variables.scss
MaratReactJs: I agree, rem is more global

MaratReactJs: move all header styles from variables.scss to header.module.scss?

MaratReactJs: removed the extra padding from the title, thanks for noticing

mw-webapp/src/component/header/Header.module.scss
```diff
@@ -5,13 +5,12 @@
...
   width: 100%;
-  padding: $header-padding;
-  background: $header-background;
+  padding: 0 1rem;
+  background: rgb(85, 122, 149);
 }
 
 .title {
   display: flex;
-  padding-top: $title-padding;
```
Howl404, line 12: No no no, title is used in many files, just pass class name from props into title and set padding 0 in it

Howl404 (changes requested):

mw-webapp/src/component/header/Header.module.scss
```diff
@@ -2,14 +2,15 @@
 
 .header {
   display: flex;
-  flex-direction: column;
+  justify-content: space-between;
+  align-items: center;
   width: 100%;
-  background: $header-background;
+  padding: 0 1rem;
```
tritonJS826, line 8: why rem?

tritonJS826 (approved):

mw-webapp/src/component/header/Header.module.scss
MaratReactJs: At first I thought of using a percentage, but for some reason I chose rem))
it's all about adaptability

Howl404 (approved):

Issue #170, opened 2023-10-29
Update models (if possible use script for add property to firebase collection)


Ekaterina1994:
Add name to Way
[ ] Update draw.io
[ ] Update models
[ ] Update service, converters, DAL if needed
[ ] Add possibility to change name if it's your way (think about move this issue to another task if will be difficult)

Howl404:
After merge will create separate issue for "Add possibility to change name if it's your way"

PR #171, opened 2023-10-29
feat: #170 update models

Howl404:


SergioAJS (approved):

mw-webapp/src/logic/waysTable/columns.tsx
```diff
@@ -21,21 +21,20 @@ export const OWNER_EMAIL = "Owner's email";
...
 export const columns = [
   columnHelper.accessor<"uuid", string>("uuid", {
-    // Maybe we should add way's name? Need to think)
     header: "Way's uuid",
 
     /**
      * Cell with owner's name
      */
     cell: ({row}) => renderCellValue(row.original.uuid),
   }),
```
Ekaterina1994, line 30: Maybe we no need render uuid of way in the table

Ekaterina1994 (changes requested):

mw-webapp/src/logic/waysTable/columns.tsx
Howl404: I think Victor said that it might be there

Ekaterina1994: If you mean this video QA session https://www.youtube.com/watch?v=fLDg2SLUcWg

Then Viktor speak about it on 2.26-2.35 .

Ekaterina1994: If you think then we will need ID then you can don't delete, but then add comment for delete it later

Howl404: Thanks, deleted

Ekaterina1994 (approved):

tritonJS826 (approved):
Yeah) that's what we need

Issue #161, opened 2023-10-23
Change links styles and logic


Ekaterina1994:
01:15 - 2:25
https://youtu.be/fLDg2SLUcWg

Ekaterina1994:
Done

PR #177, opened 2023-11-02
feat: #161 Change links styles and logic

Howl404:


mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.module.scss
```diff
@@ -0,0 +1,6 @@
+@import "src/styles/variables.scss";
+
+.link {
+  color: $lightblue;
```
Ekaterina1994, line 4: I think better to use words primary, secobdary and etc for naming of color vars. It will help us in the future to change styles/ What do you think?

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.tsx
```diff
@@ -8,6 +9,7 @@ export const renderLinkInCell = (path: string, value: string) => {
     <Link
       value={value}
       path={path}
+      className={styles.link}
```
Ekaterina1994, line 12: Maybe Im wrong but you can don't add className here because this styles will be add by default from link component?

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.tsx
```diff
@@ -8,6 +9,7 @@ export const renderLinkInCell = (path: string, value: string) => {
     <Link
       value={value}
```
Ekaterina1994, line 10: What do you think about using Link component in columns file without using additional function that just return Link component?

Ekaterina1994 (changes requested):
PLease don't forget about set reviewers )

Shustovsky (approved):

mw-webapp/src/component/link/Link.tsx
Shustovsky, line 1: I think need to add value maybe additionally the HTMLElement type

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.module.scss
Howl404: I think we can switch to primary\secondary after we have a design ready, right now it's hard to decide which color should be what

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.tsx
Howl404: I'm overriding default link styles

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.tsx
Howl404: It's a little bit more encapsulated right now, but yes we can

Ekaterina1994 (approved):

mw-webapp/src/component/link/Link.tsx
Howl404: HTMLELement inside Link? Kinda strange in my opinion

mw-webapp/src/logic/waysTable/columns.tsx
```diff
@@ -26,7 +26,13 @@ export const columns = [
     /**
      * Cell with clickable way name that leads to way page
      */
-    cell: ({row}) => renderLinkInCell(`/way/${row.original.uuid}`, row.original.name),
+    cell: ({row}) => (
+      <Link
+        path={`/way/${row.original.uuid}`}
```
tritonJS826, line 31: It's very hard to maintain template strings and it will be easy to make a bug with them. We have a file with pathes, why not to put function which returns some path there? It could be reusable in this way

mw-webapp/src/logic/waysTable/columns.tsx
```diff
@@ -45,20 +51,23 @@ export const columns = [
...
-
-    /**
-     * Cell with owner's email
-     */
-    cell: ({row}) => renderCellValue(row.original.owner.email),
+    cell: ({row}) => {
+      return (
+        <>
+          <Link
+            path={`/user/${row.original.owner.uuid}`}
```
tritonJS826, line 63: The same as for previous link

tritonJS826 (approved):
Good) minor changes required

mw-webapp/src/logic/waysTable/renderLinkInCell/renderLinkInCell.tsx
Ekaterina1994: Это приводит к неконсистентности ссылок в разных местах приложения. Поддерживать это очень сложно. Давай использовать дефолт стили для всех компонент. Например, ты можешь изменить стили в самом компоненте.  )

Issue #114, opened 2023-10-04
Use hashmap instead of find method


Ekaterina1994:
Find all .find methods  (if the array is searched more than twice) in project and use hashmap instead of it for search elements

Optimization firebase requests for WaysPreview, UserPreveiew etc

Ekaterina1994:
Done

PR #182, opened 2023-11-05
refactor: #114 Use hashmap instead of find method

Ekaterina1994:


mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
```diff
@@ -14,66 +14,47 @@ export class WayPreviewDAL {
...
+
+    return waysPreview;
+  }
+
+  /**
+   * Get WayPreview
+   */
+  public static async getWayPreview(uuid: string): Promise<WayPreview> {
+    const wayDTO = await WayService.getWayDTO(uuid);
+    const {ownerUuid, currentMentorUuids, goalUuid} = wayDTO;
```
tritonJS826, line 37: why do we need destructuring

tritonJS826 (approved):

mw-webapp/src/component/header/Header.tsx
```diff
@@ -25,26 +23,20 @@ export const Header = () => {
...
-        Hello,
-        {" "}
-        {user.displayName}
-        !
-      </h2>}
-      <div className={styles.blockButton}>
-        <Button
-          onClick={user ? logOut : logIn}
-          value={user ? BUTTON_LOG_OUT_VALUE : BUTTON_LOG_IN_VALUE}
+      <Title
```
Shustovsky, line 26: Maybe it would be better if the title was additionally designed as a link leading to the main page?

mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
Ekaterina1994: Right, no need. Deleted

mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
```diff
@@ -14,66 +14,47 @@ export class WayPreviewDAL {
...
-      //TODO: task #114 Use hashmap instead of .find
-        .find((elem) => elem.uuid === wayDTO.ownerUuid);
-      if (!ownerPreview) {
-        throw new Error(`${ownerPreview} was not found`);
-      }
-
-      return ownerPreview;
-    });
+    const waysUuids = waysDTO.map((item) => {
+      const wayUuid = item.uuid;
```
Shustovsky, line 18: can you tell me why to declare a variable here, and not immediately return the result

mw-webapp/src/component/header/Header.tsx
Ekaterina1994: As I understand this question about design, not about optimization requets?

mw-webapp/src/dataAccessLogic/WayPreviewDAL.ts
Ekaterina1994: Because I always try to return var, not just code. Do we have rules about it? If you want we can discuss about it in the group and add some rules about it))

Ekaterina1994: But yes I agree in this case we can return uuid without creating vars)

Ekaterina1994: But yes I agree in this case we can return uuid without creating vars)

Shustovsky (approved):

Issue #127, opened 2023-10-08
Add Sidebar on all pages


Howl404:
## Branch is already created

Add Sidebar(component) on all pages (maybe in header), sidebar should contain links to all other pages

Ekaterina1994:
Done

PR #186, opened 2023-11-06
feat: #127 Add Sidebar on all pages

Shustovsky:


mw-webapp/src/component/button/Button.tsx
```diff
@@ -13,7 +13,7 @@ export interface ButtonProps {
   /**
    * Callback triggered on button click
    */
-  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```
Ekaterina1994, line 16: why?

mw-webapp/src/component/header/Header.tsx
```diff
@@ -39,12 +39,45 @@ export const Header = () => {
...
           value={user ? BUTTON_LOG_OUT_VALUE : BUTTON_LOG_IN_VALUE}
         />
-        {/* //TODO: we need delete this link when after sidebar with all links to al pages will be added */}
-        <Link
-          path={pages.way.path("uuid")}
-          value={LINK_TEXT}
+        <Sidebar
+          trigger={
+            <Button value={"Navigation"} />
+          }
```
Ekaterina1994, line 45: Can we use as trigger any element? What do you think about using just text instead of button? Or it's a bad idea?

Ekaterina1994: Or we can use burgermenu for open sidebar? Dk how is better

mw-webapp/src/component/header/Header.tsx
```diff
@@ -39,12 +39,45 @@ export const Header = () => {
...
+                  <Link
+                    path={pages.userProfile.path(user.uid)}
+                    value={"Profile"}
+                  />
+                </>
+              )}
+              <Link
+                path={pages.aboutProject.path}
+                value={"About"}
+              />
```
Ekaterina1994, line 75: Can we refactor it? Looks like many similar components/ Maybe better use .map

Ekaterina1994 (changes requested):

tritonJS826 (commented):
Well done, but agreed with @Ekaterina1994 comments)

mw-webapp/src/component/header/Header.tsx
tritonJS826: Agreed, its good enough for now. Let's focus on code quality and core features.

mw-webapp/src/component/header/Header.tsx
Shustovsky: please tell me what exactly is it better for me to map?
Create a separate array of objects in this file or something else?

example:  const navigationLinks = [
        { path: pages.always.path, value: "All ways" },
         user && { path: pages.way.path(user.uid), value: "Workflow" },
         ...]

mw-webapp/src/component/button/Button.tsx
Ekaterina1994: PLease don't use onClick optionally. If you have a case where you don't need onClick you can use empty callback

mw-webapp/src/component/header/Header.tsx
Ekaterina1994: Okey, let's skeep it now

mw-webapp/src/component/header/Header.tsx
Ekaterina1994: Yea,  it will be better

Shustovsky: OK, thanks) I did it)

mw-webapp/src/component/button/Button.tsx
Shustovsky: OK) I did it

mw-webapp/src/component/header/Header.tsx
```diff
@@ -19,6 +19,33 @@ const LOGO_TEXT = "master's way";
 export const Header = () => {
   const [user, setUser] = useState<User | null>(null);
 
+  const navigationLinks = [
```
Ekaterina1994, line 22: move const in separate file

mw-webapp/src/component/header/Header.tsx
```diff
@@ -39,12 +66,28 @@ export const Header = () => {
...
+              {navigationLinks.map((item) => (
+                item && (
+                  <Link
+                    key={item.value}
+                    path={item.path}
+                    value={item.value}
+                  />
+                )
+              ))}
+            </div>
```
Ekaterina1994, line 87: move function to top of file (before return). you can call it renderNavifationLinks

Ekaterina1994 (changes requested):
Almost done)) Before approve also need to resolve conflict)

mw-webapp/src/component/header/Header.tsx
Shustovsky: I can't do this beacause some properties of the navigationLinks depend on user

mw-webapp/src/component/header/Header.tsx
Shustovsky: done

Ekaterina1994 (approved):

Howl404 (approved):

SergioAJS (approved):

Issue #168, opened 2023-10-23
Fix login (Chromium)


Ekaterina1994:
04:20 - 04:44
Uncaugh in promise FirebaseError: Firebase: Error (auth-authorized domain)
https://youtu.be/fLDg2SLUcWg

Check login in browsers: Chrome, Firefox, Opera, Safari, Edge

PR #187, opened 2023-11-07
refactor: #168 fix login chromium

SergioAJS:
1. Change signInWithRedirect method to signInWithPopup
2. Added prompt option to GoogleAuthProvider in firebase.ts file to ask the user which Google account he wants to use.

### DIFF
--- a/mw-webapp/src/service/auth/logIn.ts
+++ b/mw-webapp/src/service/auth/logIn.ts
@@ -7,18 +7,15 @@ import {auth, provider} from "src/firebase";
 export const logIn = async () => {
 
   /**
-   * TODO: IMPORTANT! Refactor required!!!
+   * TODO:
    * LogIn with signInWithPopUp and signInWithRedirect methods doesn't work in Opera.
    * Also need to check how it works in other browsers: Edge, Safari
    * Using signInWithRedirect works in Chrome but doesn't work in FireFox, Opera and maybe other browsers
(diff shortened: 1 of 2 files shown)


Ekaterina1994 (approved):

Howl404 (approved):

mw-webapp/src/service/auth/logIn.ts
```diff
@@ -1,9 +1,20 @@
...
+   * TODO: IMPORTANT!
+   * Using signInWithPopUp calls Error: Cross-Origin-Opener-Policy policy would block the window.closed call.
+   * According
+   * https://stackoverflow.com/questions/76446840/cross-origin-opener-policy-policy-would-block-the-window-closed-call-error-while
+   * The problem is with Cross-Origin-Embedder-Policy header, which is by default should be set to unsafe-none.
+   * Also Firebase recommends the next steps to resolve this issue:
+   * https://firebase.google.com/docs/auth/web/redirect-best-practices
+   *
+   */
+  await signInWithPopup(auth, provider);
```
tritonJS826, line 19: so by comment it is not clear should we change something or not .  And if yes - than why. Imagine you someone will read this comment first time, or you will read it in 1 year

tritonJS826 (approved):

mw-webapp/src/service/auth/logIn.ts
Ekaterina1994: Added comment/ Is it anough now?

Issue #188, opened 2023-11-07
Add tests


Ekaterina1994:
test cases:

1. Accordion
- [x] accordion option could be opened text visible
- [x] option could be closed (text hidden)
- [x] all options rendered
- [x] all options could be opened
- [x] all options could be opened in multiple modes
- [x] only one option could be opened in multiple mode
2. Checkbox(AKiroi)
- [x] onChange called on click and value not changed if DefaultChecked=false
- [x] onChange called on click and value changes if DefaultChecked=true
- [x] onChange callback should triggered when checkbox is clicked
3. Header(AKiroi)
- [x] alt text visible if source broken
- [x] image visible if source exists
4. Input(AKiroi)
- [x] placeholder visible
- [x] works with input mode text (some symbols are allowed and some not allowed)
- [x] works with input mode email (some symbols are allowed and some not allowed)
- [x] works with input mode number (some symbols allowed and some not allowed)
- [x] could not be used when disabled
- [x] check 'required' logic (does it work correctly)
- [ ] check how to work when props.value and props.onChange exists at the same time
5. Link
- [x] value rendered correctly
- [x] link leads to the right place
6. Modal 
- [x] modal could be opened and closed by clicking on the grey background
- [x] modal could be closed by clicking on the cross
- [x] content rendered
7. Notification
- [x] content rendered
- [x] specify what is foreground and check it works correctly
- [x] specify what is background and check it works correctly
8. Select
- [x] onChange triggered if some option selected
- [x] selected value visible
9. Sidebar
- [x] could be opened and closed by clicking on the grey background
- [x] could be closed by clicking on the cross
- [x] content rendered
10. Table
- [x] table headers rendered in the right way
- [x] data rendered in the right way
11. Title
- [x] text rendered
- [x] appropriate level exists on the page (check semantics)
12. Tooltip
- [x] content rendered
- [x] tooltip is hidden but shown on hover tooltip-trigger
13. Chart
- [ ] content rendered
14. Confirm(Alex89198900)
- [x] #628
15. Dropdown(Alex89198900)
- [x] content rendered
16. Editable text(Alex89198900)
- [ ] content rendered
17. Editable textarea(Alex89198900)
- [ ] content rendered
18. Horizontal container
- [ ] content rendered
19. Icon (test for new dev)
- [ ] content rendered
20. Image
- [ ] content rendered
21. Loader 
- [ ] content rendered
22. Progress bar(AKiroi)
- [ ] content rendered
[...]

PR #213, opened 2023-11-14
feat: #188 Add 6 tests for Accordion

dayanych:
### Overview

Implemented tests for Accordion component.

### DIFF
--- a/mw-webapp/src/component/accordion/Accordion.cy.tsx
+++ b/mw-webapp/src/component/accordion/Accordion.cy.tsx
@@ -4,29 +4,34 @@ import {AccordionTrigger} from "src/component/accordion/AccordionTrigger/Accordi
 import {getDataCy} from "src/helpers/getDataCy";
 
 describe("Accordion component", () => {
+  const FIRST_TRIGGER = "trigger1";
+  const SECOND_TRIGGER = "trigger2";
+  const FIRST_CONTENT = "content1";
+  const SECOND_CONTENT = "content2";
+
   const ACCORDION_ITEMS_EXAMPLE = [
     {
       trigger: <AccordionTrigger
-        text="Trigger1"
-        dataCy="trigger1"
+        text={FIRST_TRIGGER}
+        dataCy={FIRST_TRIGGER}
       />,
       content: <AccordionContent
         content={<div>
-          Content1
+          {FIRST_CONTENT}
         </div>}
-        dataCy="content1"
+        dataCy={FIRST_CONTENT}
       />,
     },
     {
       trigger: <AccordionTrigger
-        text="Trigger2"
-        dataCy="trigger2"
+        text={SECOND_TRIGGER}
+        dataCy={SECOND_TRIGGER}
       />,
       content: <AccordionContent
         content={<div>
-          Content2
+          {SECOND_CONTENT}
         </div>}
-        dataCy="content2"
+        dataCy={SECOND_CONTENT}
       />,
     },
   ];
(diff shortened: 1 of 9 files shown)


Ekaterina1994 (approved):

mw-webapp/src/component/accordion/Accordion.cy.tsx
```diff
@@ -0,0 +1,94 @@
...
+    cy.mount(<Accordion items={ACCORDION_ITEMS_EXAMPLE} />);
+    cy.get("[data-cy=trigger1]").click();
+
+    cy.get("[data-cy=content1]").should("be.visible");
+  });
+
+  it("should all options could be opened and closed one by one", () => {
+    cy.mount(<Accordion items={ACCORDION_ITEMS_EXAMPLE} />);
+
+    cy.get("[data-cy=trigger1]").click();
```
tritonJS826, line 58: Let's create a little helper to simplify access to elements via data-cy attributes, it could be useful when you will have different data attributes for some purposes in future
 ```const dataCy = (value:string) => `[data-cy=£{value}]`)```

tritonJS826: Let's use constants instead of strings. It is very hard to maintain strings

mw-webapp/src/component/accordion/AccordionContent/AccordionContent.tsx
```diff
@@ -18,7 +23,10 @@ export interface AccordionContentProps {
  */
 export const AccordionContent = (props: AccordionContentProps) => {
   return (
-    <RadixAccordionContent className={styles.accordionContent}>
+    <RadixAccordionContent
+      className={styles.accordionContent}
+      data-cy={props.dataCy ?? "accordion-content"}
```
tritonJS826, line 28: Great idea, but let's skip default data-cy attributes for now, at least until we will start to write e2e tests

mw-webapp/src/component/accordion/Accordion.cy.tsx
```diff
@@ -0,0 +1,94 @@
+import {Accordion, accordionTypes} from "src/component/accordion/Accordion";
+import {AccordionContent} from "src/component/accordion/AccordionContent/AccordionContent";
+import {AccordionTrigger} from "src/component/accordion/AccordionTrigger/AccordionTrigger";
+
+describe("Accordion component", () => {
+  const ACCORDION_ITEMS_EXAMPLE = [
```
tritonJS826, line 6: What do you think about little refactoring? :)  find a place in the code where accordion already used. It uses plain Json for input and helper to convert Json to appropriate format. I think we could change signature for this helper a little bit and put near the component. After this you will be able to use this helper in your test and provide clear mock Json data for the test. Also you can refactor stories a little bit)

mw-webapp/src/component/accordion/AccordionContent/AccordionContent.tsx
Ekaterina1994: In the Button component the same situation. And as I remember without default value tests don't work. But maybe I'm wrong...

mw-webapp/src/component/accordion/Accordion.cy.tsx
dayanych: Done.

mw-webapp/src/component/accordion/AccordionContent/AccordionContent.tsx
dayanych: Yes, data-cy is for tests, I need unique data.

tritonJS826: So could we get rid of default data-cy attributes in components?

mw-webapp/src/component/accordion/Accordion.cy.tsx
```diff
@@ -0,0 +1,101 @@
...
+        content={<div>
+          {SECOND_CONTENT}
+        </div>}
+        dataCy={SECOND_CONTENT}
+      />,
+    },
+  ];
+
+  it("should render the accordion and all options", () => {
+    cy.mount(<Accordion items={ACCORDION_ITEMS_EXAMPLE} />);
```
Shustovsky, line 40: DRY. I think y can create function (example mountAccordion) and use it for mount accordion

mw-webapp/src/helpers/getDataCy.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Return [data-cy=value] for CYPRESS testing
```
Shustovsky, line 2: I think you can replace [...] as data attribute

mw-webapp/src/component/accordion/AccordionContent/AccordionContent.tsx
dayanych: Done.

mw-webapp/src/component/accordion/Accordion.cy.tsx
dayanych: Done.

mw-webapp/src/helpers/getDataCy.ts
dayanych: Done.

tritonJS826 (approved):

Shustovsky (approved):

PR #214, opened 2023-11-15
Mw 180/add components to storybook

Pivodur:


tritonJS826 (approved):

mw-webapp/src/component/inProgress/InProgress.module.scss
```diff
@@ -5,6 +5,7 @@
   width: 0;
   height: 4px;
   animation: progress 1s linear forwards;
+  background-color: red;
```
Shustovsky, line 8: please use variables

Shustovsky: not only backgorund) 
please to create also a local variable of height)

Ekaterina1994 (approved):

PR #216, opened 2023-11-15
feat: #188 add tooltip tests

barrydilan:
- [x]  content rendered
- [x]  too big content (text) hidden (it seems like this functionality is not yet implemented in the project)


mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -36,7 +36,10 @@ export const Tooltip = (props: PropsWithChildren<TooltipProps>) => {
       <span className={styles.target}>
         {props.children}
       </span>
-      <div className={classes}>
+      <div
+        className={classes}
+        data-cy="tooltipContent"
```
tritonJS826, line 41: move data-cy to the variable and reuse it in the test

mw-webapp/src/component/tooltip/Tooltip.cy.tsx
```diff
@@ -0,0 +1,12 @@
+import {Tooltip} from "src/component/tooltip/Tooltip";
+
+describe("Tooltip component", () => {
+  it("Content rendered", () => {
+    const tooltipContent = "Test content";
```
tritonJS826, line 5: such constant should be declared with UPPER_CASE

tritonJS826, line 5: lets play with slightly longer text

mw-webapp/src/component/tooltip/Tooltip.tsx
Ekaterina1994: need add data-cy prop to tooltip props with optional flag.

Ekaterina1994 (changes requested):

mw-webapp/src/component/tooltip/Tooltip.tsx
tritonJS826: Agree, and then create variable in the test itself

mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -13,6 +13,10 @@
 
   &:hover + .tooltip {
     display: block;
+    overflow: hidden;
```
Ekaterina1994, line 16: why you use overflow hidden for tooltip?

barrydilan: We need overflow hidden in order to properly truncate text when its too big

mw-webapp/src/component/tooltip/Tooltip.cy.tsx
```diff
@@ -0,0 +1,26 @@
...
+    />);
+    cy.get(`[data-cy="${TOOLTIP_CONTENT_ID}"]`).should("contain.text", TOOLTIP_CONTENT);
+  });
+
+  it("Long content gets truncatet", () => {
+    cy.mount(<Tooltip
+      content={TOOLTIP_CONTENT_LONG}
+      testId={TOOLTIP_CONTENT_ID}
+    />);
+    cy.get(`[data-cy="${TOOLTIP_CONTENT_ID}"]`).invoke("show").then((text) => {
```
dayanych, line 22: We have `getDataCy`, which return data attribute for CYPRESS testing. It is better to use the function: `cy.get(getDataCy(TOOLTIP_CONTENT_ID))`

dayanych: After that, you can resolve conversation.

dayanych (approved):

mw-webapp/src/component/tooltip/Tooltip.module.scss
SergioAJS: Then how we can see the whole text?

Howl404: Yea, kinda strange, full text should be visible

tritonJS826:
abandoned

PR #218, opened 2023-11-17
feat: #211 Create Textarea component

Shustovsky:


mw-webapp/src/component/textarea/Textarea.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+  tags: ["autodocs"],
+};
+
+export default meta;
+
+/**
+ * Default story for the Textarea component.
+ */
+export const Default = () => {
+  const [text, setText] = useState<string>("");
```
tritonJS826, line 17: get rid of value (use default value instead)

mw-webapp/src/component/textarea/Textarea.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Textarea component
+ */
+export const Textarea = (props: TextareaProps) => {
+  const textareaRef = useRef<HTMLTextAreaElement | null>(null);
+
+  useEffect(() => {
+    const currentTextarea = textareaRef.current;
```
tritonJS826, line 38: why do we choose this behavior  for textarea?

mw-webapp/src/component/textarea/Textarea.stories.tsx
Shustovsky: Done

Ekaterina1994 (approved):

mw-webapp/src/component/textarea/Textarea.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <textarea
+      className={clsx(styles.textarea, props.className)}
+      placeholder={props.placeholder}
+      value={value}
+      onChange={handleTextChange}
+      rows={props.rows}
+    />
+  );
+};
```
Howl404, line 62: I think Viktar meaned to delete state and only set defaultValue to input

Howl404 (changes requested):

mw-webapp/src/component/textarea/Textarea.tsx
Shustovsky: Hmm... that's a strange remark. If Victor had something other, then he will write it by himself)))

Howl404: Do we really need to have state inside textarea? I think we will use it with onChange function 100% of the time

Howl404 (approved):

Issue #217, opened 2023-11-16
Add runtime validation to service layer (all methods)


tritonJS826:
If possible - replace dto models with zod schemas

PR #220, opened 2023-11-18
feat: #217 Add runtime validation to service layer (all methods)

Howl404:
Converted only DayReportDTO, if its ok will convert everything else too.

### DIFF
--- a/mw-webapp/src/model/DTOModel/WayDTO.ts
+++ b/mw-webapp/src/model/DTOModel/WayDTO.ts
@@ -1,57 +1,52 @@
-/**
- * Way DTO model
- */
-export class WayDTO {
+import {z} from "zod";
+
+export const WayDTOSchema = z.object({
 
   /**
    * Way's UUID
    */
-  public uuid: string;
+  uuid: z.string(),
 
   /**
    * Way's name
    */
-  public name: string;
+  name: z.string(),
 
   /**
    * @DayReport.uuids
    */
-  public dayReportUuids: string[];
+  dayReportUuids: z.array(z.string()),
 
   /**
    * Owner's UUIDs @User.uuid
    */
-  public ownerUuid: string;
+  ownerUuid: z.string(),
 
   /**
    * @MonthReport.uuids
    */
-  public monthReportUuids: string[];
+  monthReportUuids: z.array(z.string()),
 
   /**
    *Goal's UUID @Goal.uuid
    */
-  public goalUuid: string;
+  goalUuid: z.string(),
 
   /**
    * Mentor's UUIDs @User.uuid
    */
-  public currentMentorUuids: string[];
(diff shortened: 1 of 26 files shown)


tritonJS826 (commented):
Good, you can add it to every service method! just one question - what will happened if there will be invalid object?  will zod throw an Error in that case? 

By my opinion if yes - we should catch this error globally and handle it, maybe not in the current PR. Shoud I create new task for this?

tritonJS826 (approved):

Howl404:
@tritonJS826 Yea, Zod throws error with details. I will create separate issue because right now i don't know how to implement global error catcher.

mw-webapp/src/service/WayService.ts
```diff
@@ -67,9 +75,11 @@ export class WayService {
     const waysRef = collection(db, PATH_TO_WAYS_COLLECTION);
     const ownWaysQuery = query(waysRef, where("ownerUuid", "==", uuid));
     const ownWaysRaw = await getDocs(ownWaysQuery);
-    const ownWays = querySnapshotToDTOConverter<WayDTO>(ownWaysRaw);
+    const ownWaysDTO = querySnapshotToDTOConverter<WayDTO>(ownWaysRaw);
 
-    return ownWays;
+    const validatedOwnWaysDTO = WayDTOArraySchema.parse(ownWaysDTO);
+
+    return validatedOwnWaysDTO;
```
Howl404, line 82: If User doesnt have Own\Mentoring\Favorite ways, what should we do? It will throw error every time

Ekaterina1994 (approved):

tritonJS826 (approved):

mw-webapp/src/model/DTOModel/WayDTO.ts
```diff
@@ -1,57 +1,52 @@
...
-    this.ownerUuid = wayData.ownerUuid;
-    this.monthReportUuids = wayData.monthReportUuids;
-    this.goalUuid = wayData.goalUuid;
-    this.currentMentorUuids = wayData.currentMentorUuids;
-    this.isCompleted = wayData.isCompleted;
-  }
+}).strict();
 
-}
+export const WayDTOArraySchema = z.array(WayDTOSchema);
```
tritonJS826, line 47: Due to the fact that Array is a typescript type, it is usually preferable to use the word "List" in the names or simply add the letter "s", for example "WaysDTO"

tritonJS826 (approved):

mw-webapp/src/service/WayService.ts
tritonJS826: Why should it throw errors? It should be just an empty array.

Howl404: My bad, I misunderstood the errors, we should clean our DB a little bit, few ways was without name and other properties

Issue #184, opened 2023-11-06
Update table after change inputs


Ekaterina1994:
Update table§s cells after change inputs without reload page (add cells items without reload page) 
Or using state (priority)
or next way:
https://codesandbox.io/s/strange-flower-dkq5x?file=/src/App.js:5422-5434
https://github.com/TanStack/table/blob/main/examples/react/editable-data/src/main.tsx

Ekaterina1994:
Done

PR #233, opened 2023-11-22
feat: #184 Update table after change input

Ekaterina1994:


mw-webapp/src/logic/reportsTable/reportTableContext.ts
```diff
@@ -0,0 +1,31 @@
+import {createContext, useContext} from "react";
+import {DayReport} from "src/model/businessModel/DayReport";
+
+/**
+ * Dfg
```
SergioAJS, line 5: What is Dfg?

SergioAJS (commented):
Looks fine

mw-webapp/src/logic/reportsTable/reportTableContext.ts
Ekaterina1994: Deleted this file

Howl404 (approved):

SergioAJS (approved):

Issue #227, opened 2023-11-19
Add global error catcher and display notification


Howl404:
Two functions that add event listeners for error and displays notification with error
Call them inside useEffect in App

PR #246, opened 2023-11-26
feat: #227 Add global error catcher and display notification

Howl404:
Error in DayReport converter to test, open any way


mw-webapp/src/service/DayReportService.ts
```diff
@@ -35,7 +35,7 @@ export class DayReportService {
     const dayReportRaw = await getDoc(doc(db, PATH_TO_DAY_REPORTS_COLLECTION, uuid));
     const dayReportDTO = documentSnapshotToDTOConverter<DayReportDTO>(dayReportRaw);
 
-    const validatedDayReportDTO = DayReportDTOSchema.parse(dayReportDTO);
+    const validatedDayReportDTO = DayReportDTOSchema.parse({...dayReportDTO, s: 1});
```
Ekaterina1994, line 38: please explain what does s: 1 mean?

Howl404: So you can see error

tritonJS826 (commented):
I checked it in preview and it looks like notification appears only for the first error in the app. For example try to visit  wayPage few timed, first time you will see the notification, but second time you will not.

mw-webapp/src/hooks/useErrorHandler.ts
```diff
@@ -0,0 +1,34 @@
+import {useEffect, useState} from "react";
+
+/**
+ * Custom hook to handle global unhandled errors and promise rejections.
+ */
+export const useErrorHandler = () => {
+  const [error, setError] = useState("");
```
Shustovsky, line 7: please specify the type

mw-webapp/src/hooks/useErrorHandler.ts
```diff
@@ -0,0 +1,34 @@
+import {useEffect, useState} from "react";
+
+/**
+ * Custom hook to handle global unhandled errors and promise rejections.
+ */
+export const useErrorHandler = () => {
```
Shustovsky, line 6: please add what the function returns

Howl404:
> I checked it in preview and it looks like notification appears only for the first error in the app. For example try to visit wayPage few timed, first time you will see the notification, but second time you will not.

Fixed, that was happening because error message was same, so state change wasn't causing re-render

mw-webapp/src/App.tsx
```diff
@@ -1,13 +1,25 @@
...
+  const error = useErrorHandler();
 
   return (
     <>
+      {error &&
+        <Notification
+          content={<div>
+            {error}
+          </div>}
+          duration={5000}
```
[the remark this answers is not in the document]
Howl404: Removed

mw-webapp/src/hooks/useErrorHandler.ts
Howl404: Added

mw-webapp/src/hooks/useErrorHandler.ts
Howl404: Added

mw-webapp/src/hooks/useErrorHandler.ts
```diff
@@ -0,0 +1,42 @@
...
+  const [error, setError] = useState<string>("");
+
+  /**
+   * Sets a new error message, ensuring the state updates even if the message is the same
+   */
+  const setNewError = (newError: string) => {
+    setError("");
+    setTimeout(() => setError(newError), 0);
```
tritonJS826, line 14: probably we can handle it different way. Lets save an object in state instead of string. in that case solution will be scalabel (you will be able add more fields in future) and you will get rid of this workaround with resetting and resetting the state

mw-webapp/src/service/DayReportService.ts
tritonJS826: yeah, thanks for this, good idea! because of this, we was able to check behavior in preview!

mw-webapp/src/hooks/useErrorHandler.ts
Howl404: Problem is, Notification will not be re-rendered with this solution because message is the same. Only working solution that i found(except of resetting the state) is setting key Date.now() to Notification component so it will re-render

Ekaterina1994 (approved):

dayanych (approved):

tritonJS826 (approved):

Issue #228, opened 2023-11-20
Fix login in all browsers


Ekaterina1994:
logIn.ts
  /**
   * TODO:
   * LogIn with signInWithPopUp and signInWithRedirect methods doesn't work in Opera.
   * Also need to check how it works in other browsers: Edge, Safari
   * Using signInWithRedirect works in Chrome but doesn't work in FireFox, Opera and maybe other browsers
   * Using signInWithPopUp works in FireFox but calls in Chrome Error:
   * "Cross-Origin-Opener-Policy policy would block the window.closed call."
   * Resolve solutions:
   * https://firebase.google.com/docs/auth/web/redirect-best-practices
   * https://stackoverflow.com/questions/76446840/cross-origin-opener-policy-policy-would-block-the-window-closed-call-error-while
   */

PR #258, opened 2023-11-29
fix: #228 Fix login in Chrome and Firefox

Ekaterina1994:


mw-webapp/src/service/auth/writeNewUserCredentials.ts
```diff
@@ -1,21 +1,19 @@
...
  */
-export const writeNewUserCredentials = async () => {
-  const userCredentials = await getRedirectResult(auth);
-  if (!userCredentials) {
+export const writeNewUserCredentials = async (currentUser: User) => {
+  if (!currentUser) {
     return;
   }
-  if (userCredentials.user.email && userCredentials.user.displayName) {
+  if (currentUser.email && currentUser.displayName) {
```
tritonJS826, line 11: Does displayName required? If it is optional why can't we register user without it?

tritonJS826 (approved):

mw-webapp/src/service/auth/writeNewUserCredentials.ts
tritonJS826: Why do you check it than?

dayanych (approved):

mw-webapp/src/service/auth/writeNewUserCredentials.ts
Ekaterina1994: Deleted check and added default value if displayName or email is not exist

SergioAJS (approved):

PR #268, opened 2023-12-02
feat: #267 Migration to add mentorRequests to Way model

Howl404:
Don't run script


mw-migrations/src/addMentorRequests.ts
tritonJS826, line 1: lets' name the file wayMigtrations.ts and describe inside what does this migration should do. Also the nature of requests make this skript unstable (migrations could take a huge amount of time and make many changes). It's a good idea to handle errors and log result (all useful information like amount of changed models , amount of requests, full time, average request time etc. think about it, it will provide useful info for future migrations ) (for each request). In this case if something goes wrong we will be able debug it lately.

tritonJS826: also let's rename src -> migrations

tritonJS826 (approved):

Ekaterina1994 (approved):

Issue #276, opened 2023-12-05
Migrate data on firebase (becase of firebase limits)


tritonJS826:
- [x] update drawio models
- [x] add helper to get more than 30 docs in query (make multiple queries)
- [x] update data in firebase according to new models in drawio
- [x] update local models (zod)

Ekaterina1994:
Done

PR #294, opened 2023-12-09
feat: #276 Make multiple queries

Ekaterina1994:
Added multiple queries
Updated WayColumns for add ability mentors to edit cells in DayReportsTable

### DIFF
--- a/mw-webapp/src/logic/wayPage/reportsTable/WayColumns.tsx
+++ b/mw-webapp/src/logic/wayPage/reportsTable/WayColumns.tsx
@@ -29,6 +29,16 @@ const DEFAULT_SUMMARY_TIME = 0;
 const columnHelper = createColumnHelper<DayReport>();
 
 /**
+ * Get user name
+ */
+const getUserName = (users: Map<string, UserPreview>, uuid: string, name: string) => {
+  const mentor = users.get(uuid);
+  const userName = mentor ? mentor.name : name;
+
+  return userName;
+};
+
+/**
  * Columns props
  */
 interface ColumnsProps {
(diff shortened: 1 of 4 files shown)


mw-webapp/src/logic/wayPage/reportsTable/WayColumns.tsx
```diff
@@ -28,6 +28,16 @@ import styles from "src/logic/wayPage/reportsTable/WayColumns.module.scss";
...
 const columnHelper = createColumnHelper<DayReport>();
 
+/**
+ * Get user name
+ */
+const getUserName = (users: Map<string, UserPreview>, uuid: string, name: string) => {
+  const mentor = users.get(uuid);
+  const userName = mentor ? mentor.name : name;
+
+  return userName;
```
tritonJS826, line 38: what does happened here? Why to get user name we should throw it as params?

mw-webapp/src/utils/getChunkArray.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Split array on subArrays
+ */
+export const getChunksArray = <T>(array: T[], chunkLength: number): T[][] => {
+  const chunksArray = [];
```
tritonJS826, line 5: it looks here like chunks array has type any[], but it is T[][]

mw-webapp/src/utils/getChunkArray.ts
```diff
@@ -0,0 +1,12 @@
...
+
+  for (let i = 0; i < array.length; i += chunkLength) {
+    chunksArray.push(array.slice(i, i + chunkLength));
```
tritonJS826, line 8: lets create variable `chunkArray`

tritonJS826 (approved):

mw-webapp/src/utils/getChunkArray.ts
Ekaterina1994: Added type

mw-webapp/src/utils/getChunkArray.ts
Ekaterina1994: Created

dayanych (approved):

Issue #318, opened 2023-12-16
Fix button styles in reports table


SergioAJS:
![Screenshot 2023-12-16 213048](https://github.com/tritonJS826/masters-way/assets/103531486/3d9cb0b2-bad0-4f78-9562-2846a2811840)

PR #319, opened 2023-12-16
fix: #318 button styles in reports table

SergioAJS:
Screenshot 2023-12-16 213902
In WayColumns.module.scss exist class without styles. Should I delete it?
`.cell {
}`

### DIFF
--- a/mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
+++ b/mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
@@ -19,7 +19,7 @@ $listLeftMargin: 20px;
 }
 
 .flatButton {
-  width: 125px;
+  width: 100px;
   height: 20px;
   padding: 0px;
 }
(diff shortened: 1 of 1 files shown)


mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
```diff
@@ -19,7 +19,9 @@ $listLeftMargin: 20px;
 }
 
 .flatButton {
+  width: 125px;
```
tritonJS826, line 22: is default 100px width is not enough?

tritonJS826 (approved):

mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
SergioAJS: Done
Screenshot 2023-12-16 215431

mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
```diff
@@ -19,16 +19,15 @@ $listLeftMargin: 20px;
 }
 
 .flatButton {
+  width: 100px;
```
tritonJS826, line 22: Why do you need to specify width here if it is the same default width for Button component for now?

tritonJS826 (approved):

Howl404 (approved):

Issue #300, opened 2023-12-14
Add useLoad hook


tritonJS826:
maybe hook should contain next params: 
* callback to load data
* callback to validate data
* callback with success behavior
* callback with error behavior

PR #342, opened 2023-12-26
feat: #300 add useload hook

Howl404:


mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,75 @@
+import {useEffect, useState} from "react";
+
+/**
+ * Use Load hook Props
+ */
+interface useLoadProps<T, D> {
```
tritonJS826, line 6: we can use real words in generics, T and D mean nothing in the current context. try to rename them, I hope it could increase readability a lot)

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Callback this is called on fetch or validation error
+   */
+  onError?: (error: Error) => void;
+
+  /**
+   * Passed dependency to refetch and revalidate data
+   */
+  dependency?: D;
```
tritonJS826, line 31: Should dependency be kind of react state? If yes, probably it will be better to have a list of dependencies, agree?

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,75 @@
...
+        setIsLoading(true);
+        const loadedData = await loadData();
+        if (!validateData(loadedData)) {
+          throw new Error("Data validation failed");
+        }
+
+        setData(loadedData);
+        onSuccess(loadedData);
+      } catch (err) {
+        if (err instanceof Error) {
```
tritonJS826, line 63: Could you describe the case when error is not instance of Error and why we should do nothing in this case?

mw-webapp/src/logic/userPage/UserPage.tsx
```diff
@@ -60,22 +61,40 @@ export const UserPage = (props: UserPageProps) => {
...
+    loadData,
+    validateData,
+    onSuccess,
+    onError,
+    dependency: props.uuid,
+  },
+  );
 
-  if (!userPreview) {
+  if (isLoading || !userPreview) {
```
tritonJS826, line 97: do we still need to check "!userPreview"?

mw-webapp/src/logic/userPage/UserPage.tsx
```diff
@@ -60,22 +61,40 @@ export const UserPage = (props: UserPageProps) => {
...
 
-  useEffect(() => {
-    getUser();
-  }, [props.uuid]);
+  const {isLoading} = useLoad({
+    loadData,
+    validateData,
+    onSuccess,
+    onError,
+    dependency: props.uuid,
```
tritonJS826, line 93: yeah! Looks beautiful!

tritonJS826 (commented):
Why don't we use useLoad on  WayPage and UsersPage?

Howl404:
@tritonJS826 Added only to two pages(one with validation and dependency and other with simple logic) to show example of usage, will implement on other pages after hook approve

mw-webapp/src/hooks/useLoad.ts
Howl404: I don't know much about other possible Errors, i think it can be "FirebaseError" too, don't really know how to handle everyting

mw-webapp/src/logic/userPage/UserPage.tsx
Howl404: Just for TS, so userPreview is not undefined in return

mw-webapp/src/hooks/useLoad.ts
Howl404: You mean something like `dependencies?: any[]`; and pass it to useEffect?

mw-webapp/src/hooks/useLoad.ts
Howl404: Renamed

mw-webapp/src/hooks/useLoad.ts
Howl404: Refactored

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,76 @@
...
+   * Passed array of dependencies to re-fetch and re-validate data
+   */
+  dependencies: DependenciesType[];
+}
+
+/**
+ * Custom hook to fetch, validate, and manage the state of data
+ */
+export function useLoad<DataType, DependenciesType>({
+  loadData,
```
Ekaterina1994, line 38: Is it okey that hook useLoad not only fetch data but also valiadate it and other?

Howl404: We can rename it anytime, https://github.com/tritonJS826/masters-way/issues/300

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,76 @@
+import {useEffect, useState} from "react";
+
+/**
+ * Use Load hook Props
+ */
+interface useLoadProps<DataType, DependenciesType> {
```
tritonJS826, line 6: Is it conventional for us to use the postfix type? Could it improve readability in current case?

mw-webapp/src/hooks/useLoad.ts
tritonJS826: - But catch block catches all errors in js by default. So why do you want to handle only one type and why do you want to ignore errors like "FirebaseErrors"

mw-webapp/src/logic/userPage/UserPage.tsx
tritonJS826: But if is loading finished and userPreview is not exist - this case mostly looks like a bug arequired to throw a error instead of showing loader?  It's not only about typescript as I see

Howl404: This is impossible because we have onError callback on this page, i don't know how to skip userPreview check for render other way

mw-webapp/src/hooks/useLoad.ts
Howl404: I found that FirebaseError have instance of Error, so it will work

Howl404: We need to use instanceof to convert unknown into Error type, we can't pass it into onError function if we don't do it

mw-webapp/src/logic/userPage/UserPage.tsx
Howl404: Maked onError not optional

mw-webapp/src/hooks/useLoad.ts
Howl404: Removed postfix

Ekaterina1994 (approved):

SergioAJS (approved):

mw-webapp/src/logic/allWaysPage/AllWaysPage.tsx
```diff
@@ -13,16 +15,32 @@ export const AllWaysPage = () => {
...
 
-  useEffect(() => {
-    loadAllWays();
-  }, []);
+  useLoad(
+    {
+      loadData,
+      onSuccess,
+      onError,
+      dependencies: [],
```
tritonJS826, line 41: Could we make dependencies optional?

mw-webapp/src/hooks/useLoad.ts
tritonJS826: So I still don't understand why should not you handle errors if they have other type, could you explain it please?

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Passed array of dependencies to re-fetch and re-validate data
+   */
+  dependencies: Dependencies[];
+}
+
+/**
+ * Custom hook to fetch, validate, and manage the state of data
+ */
+export function useLoad<Data, Dependencies>({
```
tritonJS826, line 37: Should we use constants for consistency?

mw-webapp/src/hooks/useLoad.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Callback this is called on fetch or validation error
+   */
+  onError: (error: Error) => void;
+
+  /**
+   * Passed array of dependencies to re-fetch and re-validate data
+   */
+  dependencies: Dependencies[];
```
tritonJS826, line 31: Dependencies->Dependency

tritonJS826 (commented):
That's good, I will give an approve after implementation on all pages!

mw-webapp/src/logic/userPage/UserPage.tsx
tritonJS826: But it looks a little strange semantically for my opinion. When I see this line I think that there are two possible conditions when data is in "loading" state. Should we check all states the same way? If yes - why do we need isLoading state?

mw-webapp/src/hooks/useLoad.ts
Howl404: @tritonJS826 Any error that can occur have instance of Error, because they are all based on it, FirebaseError = Error + additional properties

mw-webapp/src/logic/userPage/UserPage.tsx
Howl404: Removed isLoading from this page

mw-webapp/src/hooks/useLoad.ts
tritonJS826: ok, if it is really safe

tritonJS826 (approved):

Issue #361, opened 2024-01-03
Add basic styles (part 2)


tritonJS826:
- [x] Fix dates import/export
- [x] add some space between interactive elements
- [x] remove horizontal scroll on hover any item
- [x] align text in tables (favorites, check other)
- [x] fast creating new way lead to missing some of them in the userPage [????]
- [x] align "Total" in dayReports table in jobs done
- [x] add redirect on create new way
- [x] align tooltip with icons (mentors block, check other)
- [x] I'm not a mentor for this way https://mastersway.netlify.app/way/Ez0jt0qOwGmn0Wsj4ju4, but it is in my mentoring table (super often regression, root cause should be fixed)
- [x] Align data in table cells mobile view (new way, short columns)
- [x] Overflow for tables (show tooltips) [???]
- [x] Get rid of accorfions (eye buttons with local storage) [6h]
- [x] check dates in all tables on mobile [3h]
- [x] check color contrast [1h] (talk with anna)
- [x] fix scroll appears on hover favorites because of tooltip [1h] (the same for comments "add" button)
- [x] jobDone, plans: aling trash icon, number [3h],
- [x] move summary in the cells to the bottom [1h]
- [x] align checkbox sizes in table [1h]
- [x] add appropriate access rules to checkbox [1h]
- [x] move check box on the line with the name, add tooltip coming soon, remove checkbox from the comments [1h]
- [x] add button - light theme - plus not visible (in Fiefox) [1h]
- [x] check tables heading in firefox (they are disappeared for me) [2h]
- [x] wide md images broke mobile layout [2h]
- [x] add gap between checkbox and description in metrics block [1h]
- [x] strange width balance in columns
- [x] wrong commentator (https://mastersway.netlify.app/way/Ez0jt0qOwGmn0Wsj4ju4) (it was me but now it is Dayana) [3h] (add migration like "formerMentors")
- [x] remove env variables from the repo
- [x] ways page: align status and waysOwner
- [x] inputs width sometimes too short, I think 3 symbols should be visible
- [x] https://mastersway.netlify.app/way/9kRSzgPVux6UYIHnhqlh cell it too large if some word could not be wrapped
- [x] can't open some ways (user with uuid is not defined (ekaterinas way))
- [x] align trash icons in mentors
- [x] Error after pressing vreate new day report
- [x] as a mentor I can't change estimations in plans for tomorrow. But I should be able
- [x] please, mark my way about read book as completed in the firebase
[...]

PR #383, opened 2024-01-14
feat: #361 align total in tables, redirect on create way, align text …

tritonJS826:
…in tables (favorites column)

### DIFF
--- a/mw-webapp/src/component/table/Table.module.scss
+++ b/mw-webapp/src/component/table/Table.module.scss
@@ -16,6 +16,21 @@ $cell-max-width: 600px;
   background: var(--primaryBackgroundColor);
   background-clip: padding-box;
   font-weight: bold;
+
+  // make createdAt fill minimum width
+  &:nth-child(1) {
+    width: 0;
+  }
+
+  // make lastUpdate fill minimum width
+  &:nth-child(2) {
+    width: 0;
+  }
+
+  // make favorites fill minimum width
+  &:nth-child(7) {
+    width: 0;
+  }
 }
 
 .table td {
--- a/mw-webapp/src/logic/wayPage/WayPage.tsx
+++ b/mw-webapp/src/logic/wayPage/WayPage.tsx
@@ -253,9 +253,11 @@ export const WayPage = (props: WayPageProps) => {
               >
                 <Button
                   value={`${Symbols.OUTLINED_STAR}${Symbols.NO_BREAK_SPACE}${favoriteForUsersAmount}`}
-                  onClick={() =>
-                    user && addFavoriteToWayAndToUser(user, way, setUser, setWay)
-                  }
+                  onClick={() => {
+                    if (user) {
+                      addFavoriteToWayAndToUser(user, way, setUser, setWay);
+                    }
+                  }}
                   buttonType={ButtonType.TERTIARY}
                 />
               </Tooltip>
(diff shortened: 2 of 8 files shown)


mw-webapp/src/logic/waysTable/waysColumns.tsx
```diff
@@ -208,14 +208,14 @@ export const waysColumns = [
      * Header
      */
     header: () => (
-      <>
+      <div>
```
SergioAJS, line 211: What purpose for using div instead of fragment?

tritonJS826: you are right, my mistake, unnecessary div removed, thanks!

SergioAJS (approved):

mw-webapp/src/component/table/Table.module.scss
```diff
@@ -16,6 +16,21 @@ $cell-max-width: 600px;
...
+    width: 0;
+  }
+
+  // make lastUpdate fill minimum width
+  &:nth-child(2) {
+    width: 0;
+  }
+
+  // make favorites fill minimum width
+  &:nth-child(7) {
```
Ekaterina1994, line 31: Maybe better to use last-child because in the future we can delte or add columns but favorite column is always at the end

mw-webapp/src/logic/wayPage/WayPage.tsx
```diff
@@ -253,9 +253,11 @@ export const WayPage = (props: WayPageProps) => {
...
                   value={`${Symbols.OUTLINED_STAR}${Symbols.NO_BREAK_SPACE}${favoriteForUsersAmount}`}
-                  onClick={() =>
-                    user && addFavoriteToWayAndToUser(user, way, setUser, setWay)
-                  }
+                  onClick={() => {
+                    if (user) {
+                      addFavoriteToWayAndToUser(user, way, setUser, setWay);
+                    }
+                  }}
                   buttonType={ButtonType.TERTIARY}
```
Ekaterina1994, line 261: why?

mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
```diff
@@ -105,12 +105,13 @@ $listLeftMargin: 20px;
...
   margin-left: $marginSmall;
-  gap: $marginSmall;
+  gap: $marginBig;
 }
 
 .list {
   display: flex;
   width: 100%;
+  min-width: 80px;
 }
```
Ekaterina1994, line 115: Shouldn't it be moved to variables?

mw-webapp/src/component/table/Table.module.scss
tritonJS826: used nth-child for consistency (3xnt looks better than first + nth + last)

mw-webapp/src/logic/wayPage/WayPage.tsx
tritonJS826: in this case function return undefined (void), before it was null if user null

mw-webapp/src/logic/wayPage/reportsTable/WayColumns.module.scss
tritonJS826: I'm not sure

Ekaterina1994 (approved):

Issue #360, opened 2024-01-03
Request optiimisation (less than 10 with no data duplicates)


tritonJS826:

tritonJS826:
Looks like wayPage still required optimisation, check please

Ekaterina1994:
Done

PR #404, opened 2024-01-17
fix: #360 Fixed bug about import nanoseconds from backup

Ekaterina1994:


mw-migrations/src/utils/getNanoSecondsThreeSymbols.ts
```diff
@@ -0,0 +1,24 @@
...
+    case 2: {
+      nanosecondsThreeSymbols = `${nanosecondsStringified}0`;
+      break;
+    }
+    default: {
+      nanosecondsThreeSymbols = nanosecondsStringified.substring(0, 3);
+      }
+  }
+
+  return nanosecondsThreeSymbols;
```
tritonJS826, line 23: improve please: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd

tritonJS826 (approved):

mw-migrations/src/utils/getNanoSecondsThreeSymbols.ts
Ekaterina1994: Done

tritonJS826 (approved):

PR #405, opened 2024-01-17
feat: #360 add optimization to requests

Ekaterina1994:


mw-webapp/src/service/WayService.ts
```diff
@@ -51,11 +54,19 @@ export class WayService {
...
+    const waysDTOQueries = chunksWaysDTO.map((chunk) => {
+      return query(waysRef, where(WAY_UUID_FIELD, "in", chunk), orderBy(WAY_CREATED_AT_FIELD, "desc"));
+    });
 
-    const validatedWaysDTO = WaysDTOSchema.parse(waysDTO);
+    const waysRaw = await Promise.all(waysDTOQueries.map(getDocs));
+    const waysDTO = querySnapshotsToDTOConverter<WayDTO>(waysRaw);
+
+    // Additional sort need because firestore method orderBy works only inside method query
+    const waysDTOOrderedByDate = waysDTO
```
tritonJS826, line 66: Do we still need this sorting?

tritonJS826 (approved):

mw-webapp/src/service/WayService.ts
Ekaterina1994: yes, if we want to sort ways

PR #406, opened 2024-01-17
fix: #360 fix dublicate requests

Ekaterina1994:


mw-webapp/src/logic/wayPage/WayPage.tsx
tritonJS826, line 1: Does everything works if you change user on this page? I mean daes data reloaded after lick on some mentor, for example?

tritonJS826 (approved):

PR #408, opened 2024-01-18
MW-403/bug null input

Ekaterina1994:


mw-webapp/src/logic/waysTable/wayStatus.tsx
```diff
@@ -3,7 +3,8 @@
  */
 enum WayStatus {
   Completed = "Completed",
-  InProgress = "In progress",
+  // It's not possible to use computed value inside string enum
+  InProgress = "In\u00A0progress",
```
tritonJS826, line 7: Yeah, would you like to use function accessor?

tritonJS826 (approved):

mw-webapp/src/logic/waysTable/wayStatus.tsx
Ekaterina1994: Updated

tritonJS826 (approved):

Issue #281, opened 2023-12-06
UserPage should be default for logged user


Ekaterina1994:

tritonJS826:
done in #165

tritonJS826:
not done

PR #413, opened 2024-01-21
feat: #281 Navigate to user page if logged in

Ekaterina1994:


mw-webapp/src/App.tsx
```diff
@@ -15,13 +17,15 @@ export const App = () => {
...
   const [user, setUser] = useState<UserPreview | null>(null);
+  const navigate = useNavigate();
 
   /**
    * OnLog in
    */
   const onLogIn = async (value: string) => {
     const currentUserPreview = await UserPreviewDAL.getUserPreview(value);
     setUser(currentUserPreview);
+    value ? navigate(pages.user.getPath({uuid: value})) : navigate(pages.allWays.getPath({}));
```
tritonJS826, line 28: What is value?

mw-webapp/src/logic/homePage/HomePage.tsx
```diff
@@ -0,0 +1,10 @@
+/**
+ * Home page
+ */
+export const HomePage = () => {
```
tritonJS826, line 4: Why do we need it?

mw-webapp/src/App.tsx
tritonJS826: Let's discuss this solution

tritonJS826 (approved):

tritonJS826 (approved):

PR #439, opened 2024-01-27
feat: #424 Add progress bar to storybook

Alex89198900:
1. Create story
2. center text in the progress bar

### DIFF
--- a/mw-webapp/src/component/progressBar/ProgressBar.module.scss
+++ b/mw-webapp/src/component/progressBar/ProgressBar.module.scss
@@ -4,6 +4,7 @@ $progressBarContainerHeight: 20px;
 $progressBarBorderRadius: 20px;
 
 .progressBarContainer {
+  position: relative;
   height: $progressBarContainerHeight;
   border-radius: $progressBarBorderRadius;
   margin-top: $marginSmall;
@@ -19,7 +20,11 @@ $progressBarBorderRadius: 20px;
 }
 
 .progressIndicatorPercent {
+  position: absolute;
+  right: 0;
+  left: 0;
   margin-right: $marginSmall;
   margin-left: $marginSmall;
   color: var(--primaryTextColor);
+  text-align: center;
 }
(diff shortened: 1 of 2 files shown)


tritonJS826 (approved):

mw-webapp/src/component/progressBar/ProgressBar.module.scss
```diff
@@ -19,7 +20,11 @@ $progressBarBorderRadius: 20px;
 }
 
 .progressIndicatorPercent {
+  position: absolute;
+  right: 0;
+  left: 0;
```
tritonJS826, line 25: how does it works at the same time?)

tritonJS826 (changes requested):

mw-webapp/src/component/progressBar/ProgressBar.module.scss
tritonJS826, line 25: why not to handle positioning with flexboxes? Flexbox css has pretty good interface

mw-webapp/src/component/progressBar/ProgressBar.module.scss
```diff
@@ -19,7 +20,11 @@ $progressBarBorderRadius: 20px;
...
   margin-right: $marginSmall;
   margin-left: $marginSmall;
   color: var(--primaryTextColor);
+  text-align: center;
```
Ekaterina1994, line 29: Please check do we need this property)

tritonJS826, line 29: text should be placed in the center of the progress bar. Does it works now this way?

mw-webapp/src/component/progressBar/ProgressBar.module.scss
Alex89198900: I can't explain exactly how it works). But this bundle makes it possible to set text-align: center.

tritonJS826: Hahah) ok I see

Ekaterina1994 (commented):
PLease rename PR accorfint to our docs

mw-webapp/src/component/progressBar/ProgressBar.module.scss
tritonJS826: so why not to use flexboxes?

tritonJS826 (approved):

Ekaterina1994 (approved):

Issue #390, opened 2024-01-15
actualize models in the docs (and firebase)


tritonJS826:
- [ ] put GoalMetrics to Goal
- [ ] update favorites for user
- [ ] add custom collections for user

PR #448, opened 2024-01-29
feat: #390 Add custom-collection-field

Ekaterina1994:


docs/DTOmodels.md
```diff
@@ -73,3 +74,8 @@ metricUuids: string[]
 description: string[]
 isDone: boolean[]
 doneDate: (timestamp|null)[]
+
+WaysCollectionDTO #collection
```
tritonJS826, line 78: why do we need another collecntion?

mw-webapp/src/dataAccessLogic/BusinessToDTOConverter/userPreviewToUserPreviewDTOPartialConverter.ts
```diff
@@ -17,6 +17,9 @@ export const userToUserDTOPartialConverter = (user: PartialWithUuid<UserPreview>
     favoriteWayUuids: user.favoriteWays,
     mentoringWayUuids: user.mentoringWays,
     createdAt: user.createdAt ? Timestamp.fromDate(user.createdAt) : undefined,
+    customWayCollectionsStringified: user.customWayCollections
+      ? user.customWayCollections.map((collection) => JSON.stringify(collection))
+      : undefined,
```
tritonJS826, line 22: why undefined? why do we need ternarian here?

docs/DTOmodels.md
Ekaterina1994: It's a subcollection/ We safw all stringified data to subcollections

PR #473, opened 2024-01-31
feat: #390 Add copiedFrom

Ekaterina1994:


mw-webapp/src/model/businessModel/Way.ts
```diff
@@ -81,6 +81,11 @@ export class Way {
    */
   public jobTags: string[];
 
+  /**
+   * Way's uuid that was copied
```
tritonJS826, line 85: from which current way was coped

tritonJS826 (approved):

Issue #472, opened 2024-01-31
Add repeat way button


tritonJS826:
Copy name, goal, metrics, tags etc. (with no reports)

PR #489, opened 2024-02-01
feat: #472 repeat the way

tritonJS826:


mw-webapp/src/dataAccessLogic/WayDAL.ts
```diff
@@ -12,13 +12,36 @@ import {DayReport} from "src/model/businessModel/DayReport";
...
+| typeof WAY_COPIED_FROM_WAY_UUID_FIELD
+| typeof WAY_GOAL_DESCRIPTION_FIELD
+| typeof WAY_ESTIMATION_TIME_FIELD
+| typeof WAY_METRICS_STRINGIFIED_FIELD
+>
+
+// {
+//   metricsStringified
+// }
+
```
Ekaterina1994, line 44: Why do we need it?

mw-webapp/src/logic/wayPage/WayPage.tsx
```diff
@@ -283,9 +283,32 @@ export const WayPage = (props: WayPageProps) => {
...
+   * Copy the way to owner
+   */
+  const repeatTheWay = async () => {
+    if (!user) {
+      throw new Error("User is not defined");
+    }
+
+    const baseWayData: BaseWayData = {
+      name: way.name,
+      copiedFromWayUuid: way.copiedFromWayUuid,
```
Ekaterina1994, line 298: I think there should ne way.uuid. Not?

Ekaterina1994 (changes requested):

Issue #388, opened 2024-01-15
adjust accessibility


tritonJS826:
Logo, user name, icons etc. (check with some plugin which use accessibility like vimium)

PR #493, opened 2024-02-02
feat: #388 Adjust accessibility 

AKiroi:
Add accessibility to logo, username
Image 1

### DIFF
--- a/mw-webapp/src/component/icon/Icon.tsx
+++ b/mw-webapp/src/component/icon/Icon.tsx
@@ -61,8 +61,8 @@ export const Icon = (props: IconProps) => {
   );
 
   return (
-    <a>
+    <button className={styles.iconWrapper}>
       {IconDictionary[props.name]({...props, className})}
-    </a>
+    </button>
   );
 };
--- a/mw-webapp/src/component/icon/Icon.tsx
+++ b/mw-webapp/src/component/icon/Icon.tsx
@@ -60,5 +60,9 @@ export const Icon = (props: IconProps) => {
     props.className,
   );
 
-  return IconDictionary[props.name]({...props, className});
+  return (
+    <a>
+      {IconDictionary[props.name]({...props, className})}
+    </a>
+  );
 };
--- a/mw-webapp/src/component/icon/Icon.tsx
+++ b/mw-webapp/src/component/icon/Icon.tsx
@@ -61,8 +61,6 @@ export const Icon = (props: IconProps) => {
   );
 
   return (
-    <button className={styles.iconWrapper}>
-      {IconDictionary[props.name]({...props, className})}
-    </button>
+    IconDictionary[props.name]({...props, className})
   );
 };
--- a/mw-webapp/src/component/icon/Icon.tsx
+++ b/mw-webapp/src/component/icon/Icon.tsx
@@ -60,7 +60,5 @@ export const Icon = (props: IconProps) => {
     props.className,
   );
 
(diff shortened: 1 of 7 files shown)


tritonJS826:
Eye icons should be interactive. Check other interactive elemets too.
eye

tritonJS826 (approved):

AKiroi:
Add accessibility to eye icons

SergioAJS (approved):

mw-webapp/src/component/icon/Icon.tsx
```diff
@@ -60,5 +60,9 @@ export const Icon = (props: IconProps) => {
     props.className,
   );
 
-  return IconDictionary[props.name]({...props, className});
+  return (
+    <a>
+      {IconDictionary[props.name]({...props, className})}
+    </a>
+  );
```
Ekaterina1994, line 67: Can you please explain why Icon is a link?

Ekaterina1994 (changes requested):

mw-webapp/src/component/icon/Icon.tsx
tritonJS826: Icon is not a link, it should be wrapped in a button

AKiroi: This is my mistake. It would have been more correct to use the <button> tag and wrap it with the Icon component. I'll fix it now.

Ekaterina1994: > Icon is not a link, it should be wrapped in a button

Icon is not always button

AKiroi: ok, i understand, i need to remove wrapper in Icon component, and create wrapper button over Icon. Fixed it in the last commit

mw-webapp/src/component/icon/Icon.tsx
```diff
@@ -60,5 +60,7 @@ export const Icon = (props: IconProps) => {
     props.className,
   );
 
-  return IconDictionary[props.name]({...props, className});
+  return (
+    IconDictionary[props.name]({...props, className})
+  );
```
Ekaterina1994, line 65: Do we need brackets?

mw-webapp/src/logic/wayPage/WayPage.tsx
```diff
@@ -454,15 +454,19 @@ export const WayPage = (props: WayPageProps) => {
...
-                  size={IconSize.MEDIUM}
-                  name="EyeOpenedIcon"
-                />
+                <button className={styles.iconWrapper}>
+                  <Icon
+                    size={IconSize.MEDIUM}
+                    name="EyeOpenedIcon"
+                  />
+                </button>
                 :
```
Ekaterina1994, line 463: Maybe will be better to use one button and change name depends on wayPageSettings.isJobDoneTagsVisible instead of render 2 buttons?

Ekaterina1994 (approved):

mw-webapp/src/logic/wayPage/WayPage.tsx
AKiroi: fixed

mw-webapp/src/component/icon/Icon.tsx
AKiroi: fixed

tritonJS826 (approved):

Issue #415, opened 2024-01-22
FIx statistics


tritonJS826:
- [ ] wrong data in the statistics block
- [x] data in the statistics block should be updated after changing data in the reports

![Image](https://github.com/tritonJS826/masters-way/assets/56690686/4a3dd62b-3857-4f48-b51b-29548382246f)

tritonJS826:
![Image](https://github.com/tritonJS826/masters-way/assets/56690686/1afd0cca-10c1-4478-8874-a72b8ec23f14)

Average time per calendar day sometime less than "Average working time in the working day"

tritonJS826:
![Image](https://github.com/tritonJS826/masters-way/assets/56690686/db15e2c0-95a3-4c0c-ad95-ada2b91512ea)

2 reports exist, but there is no data in the charts

tritonJS826:
![Image](https://github.com/tritonJS826/masters-way/assets/56690686/bda59aee-4d58-452e-a615-d052133d0c4a)

the way has only 2 reports but there is different data in different periods (should be the same, check which numbers is right)

tritonJS826:
done

EkaterinaSokolova-V:
fixed

PR #497, opened 2024-02-03
fix: #415 fix statistics

Alex89198900:
1. Days from start
2. Tags

### DIFF
--- a/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
@@ -39,7 +39,6 @@ export const MILLISECONDS_IN_DAY = 86_400_000;
  * but in application for 3 records we want to see 3 days
  */
 export const SMALL_CORRECTION_MILLISECONDS = 1;
-const CORRECTION_FOR_LAST_ELEMENT = 1;
 export const AMOUNT_DAYS_IN_WEEK = 7;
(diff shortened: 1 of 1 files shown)


mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
```diff
@@ -85,13 +85,12 @@ export const WayStatistic = (props: WayStatisticProps) => {
     return null;
   }
 
-  const allDatesTimestamps = props.dayReports.map(report => report.createdAt.getTime());
-  const maximumDateTimestamp = Math.max(...allDatesTimestamps);
-  const minimumDateTimestamp = Math.min(...allDatesTimestamps);
+  const currentDate = new Date();
+  const allReports = props.dayReports;
```
tritonJS826, line 89: there is no guarantee that the last element will be the oldest one. Also you changed the logic. Now all abandoned and completed ways  will show wrong statistics. So please base your calculations on minimum and maximum timestamps, please

tritonJS826: Also please use Math.ceil instead of Nath.round, because it will be strange to see on evening that you already spent 2 days if you just created a one report in morning.

Alex89198900: Done

PR #499, opened 2024-02-04
feat: #422 add tags to statistics

Alex89198900:
1. Add tags to statistics instead of plain text

### DIFF
--- a/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
@@ -18,11 +18,11 @@ type StatisticLineProps = {
  */
 export const StatisticLine = (params: StatisticLineProps) => {
   return (
-    <p className={styles.alignContent}>
+    <div className={styles.alignContent}>
       <span>
         {params.description}
       </span>
       {params.value}
-    </p>
+    </div>
   );
 };
--- a/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
@@ -5,17 +5,12 @@ type StatisticLineProps = {
   /**
    * Line description (left part)
    */
-  description: string;
+  description: string | React.ReactNode;
 
   /**
    * Line value (right part)
    */
   value: number | string;
-
-  /**
-   * Line component (right part)
-   */
-  component?: React.ReactNode;
 };
 
 /**
@@ -25,7 +20,7 @@ export const StatisticLine = (params: StatisticLineProps) => {
   return (
     <p className={styles.alignContent}>
       <span>
-        {params.component || params.description}
+        {params.description}
       </span>
       {params.value}
     </p>
--- a/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
@@ -11,6 +11,11 @@ type StatisticLineProps = {
    * Line value (right part)
    */
   value: number | string;
+
+  /**
+   * Line component (right part)
+   */
+  component?: React.ReactNode;
 };
 
 /**
@@ -20,7 +25,7 @@ export const StatisticLine = (params: StatisticLineProps) => {
   return (
     <p className={styles.alignContent}>
       <span>
-        {params.description}
+        {params.component || params.description}
       </span>
       {params.value}
     </p>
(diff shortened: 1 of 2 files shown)


mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
```diff
@@ -11,6 +11,11 @@ type StatisticLineProps = {
    * Line value (right part)
    */
   value: number | string;
+
+  /**
+   * Line component (right part)
+   */
+  component?: React.ReactNode;
```
tritonJS826, line 18: lets avoid to create additional props. Use polymorphism of description prop: 

```
descritpsion: string | React.ReactNode
```
code will be simplier  in this case

tritonJS826, line 18: let's avoid creating additional props. Use polymorphism of description prop: 

```
description: string | React.ReactNode
```
code will be simpler  in this case

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
```diff
@@ -22,11 +23,21 @@ export const TagStats = (props: TagStatsProps) => {
       <StatisticLine
         description={`${tagStat.name} jobs finished:`}
         value={`${tagStat.totalAmount} (${tagStat.totalAmountPercentage}%)`}
+        component={<JobTag
```
tritonJS826, line 26: Mutually exclusive properties. Will disappear after unterface refactoring.

tritonJS826 (changes requested):

Ekaterina1994 (approved):

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
Alex89198900: Done

mw-webapp/src/logic/wayPage/wayStatistics/StatisticLine.tsx
Alex89198900: Done

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
```diff
@@ -20,13 +21,23 @@ export const TagStats = (props: TagStatsProps) => {
   return props.stats.map((tagStat) => (
     <Fragment key={tagStat.name}>
       <StatisticLine
-        description={`${tagStat.name} jobs finished:`}
+        description={<JobTag
+          key={tagStat.name}
```
tritonJS826, line 25: Why do we need key here?

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
```diff
@@ -20,13 +21,23 @@ export const TagStats = (props: TagStatsProps) => {
...
+          isSmall
+        />}
         value={`${tagStat.totalAmount} (${tagStat.totalAmountPercentage}%)`}
+
       />
       <StatisticLine
         key={tagStat.name}
-        description={`${tagStat.name} time:`}
+        description={<JobTag
+          key={tagStat.name}
```
tritonJS826, line 35: And here?)

Alex89198900: And here was the key before me

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
tritonJS826: ok , check it

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
tritonJS826: ok check it too

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
Alex89198900: In this context, this is done as a warning from react: Each child in a list should have a unique "key" prop

mw-webapp/src/logic/wayPage/wayStatistics/TagStats.tsx
Alex89198900: In this context, this is done as a warning from react: Each child in a list should have a unique "key" prop

tritonJS826 (approved):

PR #500, opened 2024-02-04
feat: #496 Updated tags in DayReports

Ekaterina1994:


mw-migrations/src/migrations/dayReportsMigration.ts
```diff
@@ -35,27 +34,49 @@ const migrateDayReports = async () => {
...
-      const [jobsDone, plans, problems, comments] = await Promise.all([jobsDonePromise, plansPromise, problemsPromise, commentsPromise]);
+      const jobsDone: JobDoneDTOMigration[] = dayReport.jobsDoneStringified.map((jobDone) => JSON.parse(jobDone));
+      const plans: PlanForNextPeriodDTOMigration[] = dayReport.plansStringified.map((plan) => JSON.parse(plan));
+      const problems: CurrentProblemDTOMigration[] = dayReport.problemsStringified.map((problem) => JSON.parse(problem));
 
-      batch.set(dayReportRef,
+      const jobsDoneUpdated = jobsDone.map((jobDone) => {
+        return {
+          ...jobDone,
+          tags: jobDone.tags.length === 0
```
tritonJS826, line 45: OCreate a function and should simplify, the same everywhere

mw-webapp/src/logic/wayPage/reportsTable/modalContentJobTags/ModalContentJobTags.tsx
```diff
@@ -44,9 +46,18 @@ interface JobDoneTagsProps {
...
+  const filteredJobTags = isJobTagsSingle
+    ? jobTagsUpdated
+    : Array.from(new Set(jobTagsUpdated)).filter((tag) => tag !== "no tag");
+
+  const checkedJobTags = isJobTagsEmpty
+    ? jobTagsUpdated.concat("no tag")
+    : filteredJobTags;
 
-  const allTags = Array.from(new Set(props.jobTags.concat(props.jobDoneTags).filter((tag) => tag !== DEFAULT_TAG)));
+  const allTags = Array.from(new Set(props.jobTags.concat(checkedJobTags).filter((tag) => tag !== DEFAULT_TAG)));
```
tritonJS826, line 60: Why do we need set here?

tritonJS826 (approved):

mw-webapp/src/logic/wayPage/reportsTable/modalContentJobTags/ModalContentJobTags.tsx
Ekaterina1994: Because we dont want to see tags duplicates

mw-migrations/src/migrations/dayReportsMigration.ts
Ekaterina1994: I don't think that it is needed for migration

tritonJS826: Yeah, rigth

mw-webapp/src/logic/wayPage/reportsTable/modalContentJobTags/ModalContentJobTags.tsx
tritonJS826: Whu should we have duplicates?

Ekaterina1994: Because props.jobTags is all tags on the way and checkedJobTags is tags that was used on special joDone or plan. So all jobsDoneTags includes checkedJobTags by default, but we don't know which one. So for delete duplicate we use Set

Issue #429, opened 2024-01-25
Job tags improvements


tritonJS826:
- [x] make tags clicable in the modal 
- [x] create the tag on press enter in modal
- [x] add placeholder in the modal if there is no tag available
- [x] ellipsis tag (play with max length (img 1)
- [ ] align tags in the cells vertically (max input 9999) (шьп 3)
- [x] add primary button in modals
- [x] add cancel button in modals

![Image](https://github.com/tritonJS826/masters-way/assets/56690686/9b797eca-7077-443d-ab85-07fc6336a95e)

![Image](https://github.com/tritonJS826/masters-way/assets/56690686/0151d688-9aa9-445e-a824-21a2ceb3440d)

![Image](https://github.com/tritonJS826/masters-way/assets/56690686/ab2c981c-bea0-41c2-a793-0c1dda649840)

tritonJS826:
Tags should be aligned vertically

PR #503, opened 2024-02-05
feat: #429 Fixed tags styles, update status way, expand mentors rights

Ekaterina1994:


mw-webapp/src/logic/wayPage/goalBlock/GoalBlock.tsx
```diff
@@ -60,9 +60,13 @@ export const GoalBlock = (props: GoalBlockProps) => {
    * Update goal
    */
   const updateGoalMetrics = async (metricsToUpdate: Metric[]) => {
+    const metricsStatus = metricsToUpdate.map((metric) => metric.isDone);
```
tritonJS826, line 63: If metricsStatus doe not used in other places you can use method .some or .each

tritonJS826 (approved):

tritonJS826 (approved):

Issue #419, opened 2024-01-23
Formatting numbers in plans and jobs


tritonJS826:
Please remove first character "0" when typing in input number (use .parseInt method)

PR #515, opened 2024-02-08
fix: #419 formatting numbers in plans and jobs

Alex89198900:
1. Formatting numbers in plans and jobs


mw-webapp/src/component/input/Input.tsx
```diff
@@ -69,16 +69,19 @@ interface InputProps {
...
    * Event handler for the input change event
    */
   const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     props.onChange(event.target.value);
+    refInput.current.value = "";
   };
 
   return (
     <input
+      ref={refInput}
```
tritonJS826, line 84: Does it possible to solve this issue with no refs?

Alex89198900: I tried to solve this using the parseInt method, but the fact is that when we focus, we have the values ​​in the input from the database, and when you enter they are simply added since the entered values ​​have not yet entered the database.

Ekaterina1994: This link should help with this issue

https://github.com/facebook/react/issues/9402

mw-webapp/src/utils/FormatterUtils.ts
```diff
@@ -0,0 +1,15 @@
+const FIRST_CHAR = 1;
+
```
Ekaterina1994, line 2: I think it's not clear naming

mw-webapp/src/utils/FormatterUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Formatter utils
+ */
+export class FormatterUtils {
+
+  /**
+   * Checking for zero and formatting the value
+   */
+  public static defaultFormatter(value: string) {
+    return value.startsWith("0") ? value.slice(FIRST_CHAR) : Number(value);
+  }
```
Ekaterina1994, line 13: Do we really need a special util , maybe better to use line 12 in the EditableText on line 76? Let's discuss how will be better)

mw-webapp/src/utils/FormatterUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Formatter utils
+ */
+export class FormatterUtils {
+
+  /**
+   * Checking for zero and formatting the value
+   */
+  public static defaultFormatter(value: string) {
```
Ekaterina1994, line 11: Bad naming

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -72,7 +73,7 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
    * Check type of coming value and convert it to Number if need to use input with type "number"
    */
   const setValue = (value: string) => {
-    const number = Number(value) ?? 0;
+    const number = FormatterUtils.defaultFormatter(value);
```
Ekaterina1994, line 76: method defaultFormatter return number or string but const naming is number. I think we need to change naming or update method for return only number

Ekaterina1994 (changes requested):

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -72,16 +73,29 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
    * Check type of coming value and convert it to Number if need to use input with type "number"
    */
   const setValue = (value: string) => {
-    const number = Number(value) ?? 0;
-    const updatedValue = props.type === "number" ? number : value;
-    setText(updatedValue as T);
+    const checkedValue = value || 0;
+    setText(checkedValue as T);
```
tritonJS826, line 77: пget rid of as operator

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -72,16 +73,29 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
...
+  };
+
+  /**
+   * Formatting value entered into the input
+   */
+  const formatterValue = (value: string|number) => {
+    const formattedValue = FormatterInputValue.defaultFormatter(value);
+    const updatedValue = props.type === "number" ? formattedValue : value;
+    useEffect(() => {
```
tritonJS826, line 86: formatterValue is not a hook. You can't execute useEffect in the plain function

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -72,16 +73,29 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
    * Check type of coming value and convert it to Number if need to use input with type "number"
    */
   const setValue = (value: string) => {
-    const number = Number(value) ?? 0;
-    const updatedValue = props.type === "number" ? number : value;
-    setText(updatedValue as T);
+    const checkedValue = value || 0;
```
tritonJS826, line 76: this looks a little strange, check it

mw-webapp/src/component/editableText/EditableText.tsx
Alex89198900: A simple function will not solve this problem
"EditableText.tsx:84 Warning: Cannot update a component (`EditableText`) while rendering a different component (`Input`). To locate the bad setState() call inside `Input`, follow the stack trace as described in"

mw-webapp/src/component/editableText/EditableText.tsx
Alex89198900: This is a check to ensure that the state does not contain an empty string.

mw-webapp/src/component/editableText/EditableText.tsx
Alex89198900: This is checking for an empty string

mw-webapp/src/component/input/Input.tsx
```diff
@@ -61,25 +62,51 @@ interface InputProps {
...
+    let parsedValue: T;
+
+    switch (props.type) {
+      case "number": {
+        parsedValue = props.parser
+          ? props.parser(event.target.value)
+          : ParserInputValue.defaultNumberParser(event.target.value);
+        break;
+      }
+      default: {
```
tritonJS826, line 97: !

mw-webapp/src/component/input/formatters.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Formatters
+ */
+export class FormatterInputValue {
+
+  /**
+   * Delete first zero from input number
+   */
+  public static withNoFirstZero(value: string | number) {
```
tritonJS826, line 9: decompose

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -69,24 +70,23 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
...
   /**
    * Render input
    */
   const renderInput = () => (
     <Input
+      formatter={FormatterInputValue.withNoFirstZero}
       type={props.type ?? "text"}
       max={props.max}
-      value={text}
+      value={value}
```
tritonJS826, line 87: use type guard for separating two different cases for formater string and number

tritonJS826 (approved):

tritonJS826 (approved):

mw-webapp/src/component/input/formatters.ts
tritonJS826: So how does it work for now?

mw-webapp/src/component/editableText/EditableText.tsx
Ekaterina1994: Used

mw-webapp/src/component/input/formatters.ts
Ekaterina1994: Done

mw-webapp/src/component/input/formatters.ts
```diff
@@ -0,0 +1,28 @@
...
+   */
+  public static withNoFirstZero(value: number) {
+    const SECOND_INDEX = 1;
+    const valueStringified = value.toString();
+
+    const formattedValue = valueStringified.startsWith("0")
+      ? valueStringified.slice(SECOND_INDEX)
+      : value;
+
+    return formattedValue;
```
tritonJS826, line 17: This formatter is complex a little. Why not to just use Number(value)?

tritonJS826 (approved):

Ekaterina1994 (approved):

tritonJS826 (approved):

PR #524, opened 2024-02-11
feat: #506 add delay 1s to tooltip

AKiroi:
add delay 1sec to tooltip


mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -47,3 +49,15 @@ th .wrapper {
...
+  0% {
+    opacity: 0;
+  }
+  50% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
```
tritonJS826, line 63: Yeah, that's beautiful! I like this solution

tritonJS826 (approved):

Issue #560, opened 2024-02-21
Filters does not works


tritonJS826:
completed ways are not rendered at all. I assume that way with 100% progress is completed.
the issue persists both at user's page and ways page

PR #563, opened 2024-02-22
fix: #560 Fix filters and links

Ekaterina1994:


mw-webapp/src/component/link/Link.tsx
```diff
@@ -26,13 +27,14 @@ interface LinkProps {
...
  */
-export const Link = (props: LinkProps) => {
+export const Link = (props: PropsWithChildren<LinkProps>) => {
   return (
     <LinkFromRouter
       className={clsx(styles.link, props.className)}
       to={props.path}
     >
       {props.value}
+      {props.children}
```
tritonJS826, line 37: Чwhy do we need both value and children? What the difference?

tritonJS826 (approved):

mw-webapp/src/component/link/Link.tsx
Ekaterina1994: Yea, thank you, deleted

SergioAJS (approved):

Issue #533, opened 2024-02-13
Editing time in jobs should not change layout


tritonJS826:

tritonJS826:
done

PR #595, opened 2024-03-03
style: #533 editing time in jobs should not change layout

Alex89198900:
1. Editing time in jobs should not change layout

### DIFF
--- a/mw-webapp/src/component/editableText/EditableText.module.scss
+++ b/mw-webapp/src/component/editableText/EditableText.module.scss
@@ -6,8 +6,8 @@ $editableTextMinWidth: 40px;
 .editableText {
   overflow: hidden;
   width: $editableTextWidth;
-
-  input {
-    min-width: $editableTextMinWidth;
+  input[type="number"]::-webkit-inner-spin-button {
+    margin: 0;
+    -webkit-appearance: none;
   }
 }
--- a/mw-webapp/src/component/editableText/EditableText.module.scss
+++ b/mw-webapp/src/component/editableText/EditableText.module.scss
@@ -6,8 +6,15 @@ $editableTextMinWidth: 40px;
 .editableText {
   overflow: hidden;
   width: $editableTextWidth;
-  input[type="number"]::-webkit-inner-spin-button {
-    margin: 0;
-    -webkit-appearance: none;
-  }
+}
+
+.inputEditableText::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+}
+
+.inputEditableText {
+  width: 30px;
+  -moz-appearance: textfield;
+  appearance: textfield;
+  text-align: center;
 }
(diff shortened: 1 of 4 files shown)


mw-webapp/src/component/editableText/EditableText.module.scss
```diff
@@ -6,8 +6,8 @@ $editableTextMinWidth: 40px;
 .editableText {
   overflow: hidden;
   width: $editableTextWidth;
-
-  input {
-    min-width: $editableTextMinWidth;
+  input[type="number"]::-webkit-inner-spin-button {
+    margin: 0;
+    -webkit-appearance: none;
   }
```
Ekaterina1994, line 12: 1. Could you please check is it work in Firefox?
2. Is it possible to use className and add -webkit-inner-spin-button  instead of adding property to all number inputs?
3. Can you please explain difference between  -webkit-inner-spin-button  and -webkit-outer-spin-button

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -96,6 +96,7 @@ export const EditableValue = <T extends string | number>(props: EditableTextProp
       value={value}
       autoFocus={true}
       onChange={updateValue}
+      style={{width: String(value).length + "ch"}}
     />
```
Ekaterina1994, line 100: What do you think about to set the width to always be the same (equal to 4 characters, but in pixels)?

mw-webapp/src/component/input/Input.tsx
```diff
@@ -74,6 +74,12 @@ interface InputProps<T extends string | number> {
    */
   parser?: (value: string) => T;
 
+  /**
+   * Style for input
+   */
+  style?: {
+    [key: string]: string;
+  };
 }
```
Ekaterina1994, line 83: I think it's not the best idea to use additional props style. Let's discuss about it in the group? Can you please write there advantages and disadvantages of this method? )

mw-webapp/src/component/editableText/EditableText.module.scss
Alex89198900: 1. Done

Alex89198900: 2. Done

Alex89198900: 3. No

mw-webapp/src/component/editableText/EditableText.tsx
Alex89198900: Done

mw-webapp/src/component/input/Input.tsx
Alex89198900: I removed props style and did as you advised

tritonJS826:
Done

Issue #428, opened 2024-01-24
Add basics styles (part 3)


tritonJS826:
- [ ] waypage: fix tooltip on Date, fix tooltip on numbers in the table (time), align margins for all tooltips (add comment, add problem, check everywhere, metrics block etc.). Align in the whole app
- [x] statistics block does not react on change job time 
- [ ] move goal tooltip above the goal (for example on ways page)
- [ ] sorry, i added short columns for all table in project, could you apply this style only for appoipriate columns 
- [ ] wrong numbers in statistics block (for example totoa week and total 2weeks) check all, probably wrong intervals [2h]
- [ ] add tooltip for all interactive elements, adjust all tooltips position (in the whole app, check twice it is very important) (add all 8 positions and use an appropriate)
- [ ] remove not required wrappers, learn flex box API, align goal singular metric

tritonJS826:
https://github.com/tritonJS826/masters-way/assets/56690686/c626acfa-c8f7-452a-97e9-c22f04a418d0

PR #613, opened 2024-03-05
fix: #428 fixed tooltip in app

AKiroi:
- [x] waypage: fix tooltip on Date, fix tooltip on numbers in the table (time), align margins for all tooltips (add comment, add problem, check everywhere, metrics block etc.). Align in the whole app
- [x] statistics block does not react on change job time 
- [x] move goal tooltip above the goal (for example on ways page)
- [x] sorry, i added short columns for all table in project, could you apply this style only for appoipriate columns 
- [ ] wrong numbers in statistics block (for example totoa week and total 2weeks) check all, probably wrong intervals [2h]
- [x] add tooltip for all interactive elements, adjust all tooltips position (in the whole app, check twice it is very important) (add all 8 positions and use an appropriate)
- [ ] remove not required wrappers, learn flex box API, align goal singular metric

### DIFF
--- a/mw-webapp/src/component/tooltip/Tooltip.module.scss
+++ b/mw-webapp/src/component/tooltip/Tooltip.module.scss
@@ -35,19 +35,32 @@ th .wrapper {
 }
 
 .right {
-  transform: translateY(-100%) translateX($tooltipMarginLeft);
-}
-
-// Don't delete empty style because it's one of default tooltip position className and can be change in the future
-.bottom {
+  transform: translateY(-100%) translateX($marginBig);
 }
 
 .left {
   transform: translateY(-100%) translateX(calc(-100% - $marginBig));
 }
 
+// Don't delete empty style because it's one of default tooltip position className and can be change in the future
+.bottom {
+}
+
+.bottomLeft {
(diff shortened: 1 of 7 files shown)


tritonJS826 (approved):

mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -36,21 +36,34 @@ th .wrapper {
...
 
-.left {
-  transform: translateY(-100%) translateX(calc(-100% - $marginBig));
+.bottomLeft {
+  transform: translateY(0) translateX(-50%);
+}
+
+.bottomRight {
+  transform: translateY(0) translateX(50%);
+  // transform: translateY(calc(0% - $marginBig)) translateX(50%);
```
Ekaterina1994, line 56: Do you need this comment?

Ekaterina1994 (approved):

mw-webapp/src/component/tooltip/Tooltip.module.scss
AKiroi: fixed

PR #644, opened 2024-03-20
feat: #188 add test to dropdown

Alex89198900:
1. Add test to dropdown

### DIFF
--- a/mw-webapp/src/component/dropdown/Dropdown.cy.tsx
+++ b/mw-webapp/src/component/dropdown/Dropdown.cy.tsx
@@ -1,0 +1,67 @@
+import {Button} from "src/component/button/Button";
+import {Dropdown} from "src/component/dropdown/Dropdown";
+import {getDataCy} from "src/utils/cyTesting/getDataCy";
+
+const DROPDOWN_CY = {
+  dataCyTrigger: "trigger",
+  dataCyOverlay: "overlay",
+  dataCyContent: "content",
+};
+
+describe("Dropdown component", () => {
+
+  beforeEach(() => {
+    cy.mount(
+      <Dropdown
+        cy={DROPDOWN_CY}
+        trigger={
+          <Button
+            value="DROPDOWN"
+            onClick={() => {}}
+            dataCy={DROPDOWN_CY.dataCyTrigger}
+          />
+        }
+        dropdownMenuItems={[
+          {
+            id: "#1",
+            value: "tes#1",
+
+            /**
+             * Click action
+             */
+            onClick: () => {},
+          },
+          {
+            id: "#2",
+            value: "tes#2",
+
+            /**
+             * Click action
+             */
+            onClick: () => {},
+          },
+        ]}
+      />,
+    );
+  });
+
+  it("should not exist by default", () => {
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContent))
+      .should("not.exist");
+  });
+  it("should close by clicking on background", () => {
+    cy.get(getDataCy(DROPDOWN_CY.dataCyTrigger))
+      .click();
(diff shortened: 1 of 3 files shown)


mw-webapp/src/component/dropdown/Dropdown.cy.tsx
```diff
@@ -0,0 +1,67 @@
...
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContent))
+      .should("not.exist");
+  });
+
+  it("should render content", () => {
+    cy.get(getDataCy(DROPDOWN_CY.dataCyTrigger))
+      .click();
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContent))
+      .should("exist");
+  });
```
Ekaterina1994, line 66: Also you can check
1. is options correct. (length and that each option render right value
2. Should select right option after click on value

tritonJS826 (approved):

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
Alex89198900: Done

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  it("should length and that each option render right value", () => {
+    cy.get(getDataCy(DROPDOWN_CY.dataCyTrigger))
+      .click();
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContentList)).each(($ele) => {
+      if ($ele.text() === "tes#1") {
```
Ekaterina1994, line 75: Can you please explain what does $ele. mean?

tritonJS826: for js it is not convenient to start name variables with "$" sign) just name it properly (for example "dropdownOption" etc.)

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
```diff
@@ -0,0 +1,92 @@
...
+        cy.wrap($ele).click();
+      }
+    });
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContentList)).each(($ele) => {
+      if ($ele.text() === "tes#2") {
+        cy.wrap($ele).click();
+      }
```
tritonJS826, line 82: you can get element by text in cypress

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
```diff
@@ -0,0 +1,92 @@
...
+    });
+    cy.get(getDataCy(DROPDOWN_CY.dataCyContentList)).find("li").should("have.length", DROPDOWN_LENGHTH);
+  });
+
+  it("should select right option after click on value", () => {
+    cy.get(getDataCy(DROPDOWN_CY.dataCyTrigger)).click();
+    cy.contains(getDataCy(DROPDOWN_CY.dataCyContentList), "tes#1").click();
+    cy.contains(getDataCy(DROPDOWN_CY.dataCyContentList), "tes#2").click();
```
tritonJS826, line 90: if "tes" mean "test"  - lets use the full word

tritonJS826 (approved):

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
Alex89198900: Done

mw-webapp/src/component/dropdown/Dropdown.cy.tsx
Alex89198900: Done

Ekaterina1994 (approved):

PR #647, opened 2024-03-22
Mw 188/create tests image loader

AKiroi:
- create tests to Icon Component
- create tests to Loader Component

### DIFF
--- a/mw-webapp/src/component/image/Image.cy.tsx
+++ b/mw-webapp/src/component/image/Image.cy.tsx
@@ -4,33 +4,51 @@ import {Image} from "src/component/image/Image";
 import {getDataCy} from "src/utils/cyTesting/getDataCy";
 
 const IMAGE_CY = "image";
+const WRONG_SRC = "./wrong.src";
+
+/**
+ * CreateTestImageProps props
+ */
+export interface createTestImageProps {
+
+  /**
+   * Image's src
+   */
+  src: string;
+
+}
+
+/**
+ *Create test input component
+ */
+const createTestImage = (props: createTestImageProps) => {
+  return (
+    <Image
+      dataCy={IMAGE_CY}
+      src={props.src}
+      alt={LOGO_TEXT}
+    />
+  );
+};
 
 describe("Image component", () => {
 
-  beforeEach(() => {
-    cy.mount(
-      <Image
-        dataCy={IMAGE_CY}
-        src={logo}
-        alt={LOGO_TEXT}
-      />,
-    );
-  });
-
   it("should render image component correctly", () => {
+    cy.mount(createTestImage({src: logo}));
     cy.get(getDataCy(IMAGE_CY)).should("exist");
   });
 
-  it("should display a default image if src is not provided", () => {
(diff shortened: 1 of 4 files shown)


mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -0,0 +1,36 @@
...
+      />,
+    );
+  });
+
+  it("should render image component correctly", () => {
+    cy.get(getDataCy(IMAGE_CY)).should("exist");
+  });
+
+  it("should display a default image if src is not provided", () => {
+    cy.get(getDataCy(IMAGE_CY)).should("have.attr", "src").and("not.be.empty");
```
Ekaterina1994, line 25: Please explain what is default image should render if src is not provided? Is it our image?

mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it("should display the image with correct alt text", () => {
+    cy.get(getDataCy(IMAGE_CY)).should("have.attr", "alt", LOGO_TEXT);
+  });
+
+  it("image component should be visible", () => {
+    cy.get(getDataCy("image")).should("be.visible");
```
Ekaterina1994, line 33: PLease tell me about difference .should("exist") and .should("be.visible") methods)

mw-webapp/src/component/loader/Loader.cy.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  it("should have the alt text", () => {
+    cy.get(getDataCy(LOADER_CY)).find("img").should("have.attr", "alt", ALT_TEXT);
+  });
+
+  it("should have the correct src attribute for the image", () => {
+    cy.get(getDataCy(LOADER_CY))
+      .find("img")
+      .should("have.attr", "src")
+      .and("include", "mastersWayLogo.svg");
```
Ekaterina1994, line 32: If we will change Logo image test will be broke

mw-webapp/src/component/image/Image.cy.tsx
AKiroi: The main difference between these checks is that the first one verifies only the presence of the element in the DOM, while the second one verifies the visibility of the element on the page.

mw-webapp/src/component/loader/Loader.cy.tsx
AKiroi: yes, the test will be broke. Maybe it not a good test, removed it

mw-webapp/src/component/image/Image.cy.tsx
AKiroi: it's not a correxct test, i want to create test like "displays the alt text if the image is not displayed", removed it, anr create a new test "displays the alt text if the image is not displayed

tritonJS826 (approved):
merge after Ekaterina's approve

Ekaterina1994 (approved):

PR #678, opened 2024-04-14
feat: #188 add test icon

Alex89198900:
1.  Add test icon


tritonJS826:
wrong branch should be MW-622/add-service layer

PR #720, opened 2024-04-21
GitHub issue template

pp2amd6:


Ekaterina1994:
Why do we need this?

pp2amd6:
> Why do we need this?

The template defines the layout for the report: presets blocks, specifies labels.

A single report template reduces the time for creating a report, because the necessary mandatory blocks for a bug report are ready. The template brings bug reports to a single form, which simplifies working with them, regardless of who created it or who will fix it.

As a tester, I won’t need to specify the required attributes of a bug report over and over again, because... this will already be written down.

When creating a report, the user can also choose not to use a template.

tritonJS826:
Lets discuss it together)

tritonJS826 (changes requested):
too much changes. Please recheck

pp2amd6:
Changed in https://github.com/tritonJS826/masters-way/pull/722

Issue #723, opened 2024-04-21
The click outside of "Add tag" button call new tag popup


pp2amd6:
The click outside of "Add tag" button call new tag popup

## Precondition
1. The user is already authorized in the application using his Google account
2. The user has a mentee
3. Mentee way page already opened

## Steps
1. Click on "Adjust job tags" button
2. Click outside of "Add tag"  button
3. Popup window for entering tag name is shown
![Image](https://github.com/tritonJS826/masters-way/assets/4976837/6d7d02b6-ad7b-4b86-9bd9-747da5770477)

## Actual result
The click outside of "Add tag" call create new tag popup

## Expected result
The click outside of "Add tag" does not cause any reaction

tritonJS826:
https://github.com/tritonJS826/masters-way/assets/56690686/d968b20a-38ca-4f0a-b012-46b3f474d956

PR #806, opened 2024-05-02
fix: #723 fixed mistake with Add tag button

AKiroi:
resolved problem in this task https://github.com/tritonJS826/masters-way/issues/723

https://github.com/tritonJS826/masters-way/assets/112805668/a4f92256-3639-44c6-ad15-26c04ce68c1c

### DIFF
--- a/mw-webapp/src/logic/wayPage/labels/AdjustLabelsModalContent.tsx
+++ b/mw-webapp/src/logic/wayPage/labels/AdjustLabelsModalContent.tsx
@@ -106,24 +106,27 @@ export const AdjustLabelsBlock = (props: JobTagsProps) => {
       }
 
       {props.isEditable &&
-        <Modal
-          isOpen={isJobDoneModalOpen}
-          content={
-            <PromptModalContent
-              placeholder={LanguageService.way.filterBlock.jobTagPlaceholder[language]}
-              close={() => setIsJobDoneModalOpen(false)}
-              onOk={createJobTag}
-              okButtonValue={LanguageService.modals.promptModal.okButton[language]}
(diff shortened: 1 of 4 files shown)


tritonJS826 (approved):

tritonJS826:
https://github.com/tritonJS826/masters-way/assets/56690686/64338711-888f-48a3-915b-cee4360989d3

mw-webapp/src/logic/wayPage/labels/AdjustLabelsModalContent.tsx
```diff
@@ -106,24 +106,27 @@ export const AdjustLabelsBlock = (props: JobTagsProps) => {
...
-            />
-          }
-          trigger={
-            <Button
-              value={LanguageService.way.filterBlock.addTagButton[language]}
-              onClick={() => setIsJobDoneModalOpen(true)}
-            />
-          }
-        />}
+        <div className={styles.wrapperButton}>
```
Ekaterina1994, line 109: 1. Bad naming, if name wrapperButton this div should be around button but not around all modal
2. I thimk that better to add this class inside modal component for get rid the same situation in the future

@tritonJS826  What do you think?

Ekaterina1994 (changes requested):
Wrote comment

mw-webapp/src/logic/wayPage/labels/AdjustLabelsModalContent.tsx
AKiroi: fixed

tritonJS826 (approved):

Issue #396, opened 2024-01-15
Make image component clickable (fullscreen image)


tritonJS826:

Ekaterina1994:
@AKiroi  This issue is done yet  (feat: #396 Add option to open image in full screen #970) 
Please finish PR (some minor changes needed in tests)

Ekaterina1994:
Done

vdruda:
Done

PR #812, opened 2024-05-03
feat: #396 added option to open Image in full screen

FriendOfEveryone:
disabled husky pre-commit 
feat: #396 added option to open Image in full screen


mw-webapp/pnpm-lock.yaml
Ekaterina1994, line 1: What and why has changed in this file?

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,15 @@
+.imageWrapperFullScreen {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  width: 100vw;
+  height: 100vh;
+  padding: 7%;
```
Ekaterina1994, line 9: don't use %

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
   return (
-    <img
-      src={props.src}
-      alt={props.alt}
-      className={className}
-      data-cy={props.dataCy}
-    />
+    <div className={isOpen ? styles.imageWrapperFullScreen : ""}>
+      <img
+        tabIndex={0}
```
Ekaterina1994, line 55: why do we need this property?

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,15 @@
+.imageWrapperFullScreen {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
Ekaterina1994, line 6: Don't enough only top and left for position image?

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,15 @@
...
+  width: 100vw;
+  height: 100vh;
```
Ekaterina1994, line 8: Isn't is too big image?

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
    * Data attribute for cypress testing
    */
   dataCy?: string;
+
+  /**
+   * Should image open?
```
Ekaterina1994, line 31: Delete "?" at the end

Add default value

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
  * Сomponent for displaying images
  */
 export const Image = (props: ImageProps) => {
+
+  const [isOpen, setIsOpen] = useState(false);
+
   const className = clsx(styles.image, props.className);
 
+  /**
+   * Function full screen image
```
Ekaterina1994, line 46: Not clear comment

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
+   */
+  const onClickHandler = (value: boolean) => {
+    setIsOpen(value);
+  };
```
Ekaterina1994, line 50: Maybe better don't use additional wrapper and use setIsOpen inside code instead of function-wrapper? What do you think?

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
+
   return (
-    <img
-      src={props.src}
-      alt={props.alt}
-      className={className}
-      data-cy={props.dataCy}
-    />
+    <div className={isOpen ? styles.imageWrapperFullScreen : ""}>
```
Ekaterina1994, line 53:
```suggestion
    <div className={isOpen ?? styles.imageWrapperFullScreen}>
```

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
+      <img
+        tabIndex={0}
+        src={props.src}
+        alt={props.alt}
+        className={className}
+        data-cy={props.dataCy}
+        onClick={props.isShouldOpen ? () => onClickHandler(true) : undefined}
+        onBlur={() => onClickHandler(false)}
```
Ekaterina1994, line 61: Maybe better to use Modal and render big image after click by trigger. 

@tritonJS826 What do you think?

If not, then need add possibility to close image not only by onBlur - it can be not clear for user

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: Which? tabIndex?

mw-webapp/src/component/image/Image.module.scss
FriendOfEveryone: depending on the orientation of the picture this will change

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: Ok

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: Ok. I will change

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: ok

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: I can add a transparent button in the corner with X

mw-webapp/pnpm-lock.yaml
FriendOfEveryone: I didn't change anything manually. Reinstalled packages and modules

mw-webapp/src/component/image/Image.module.scss
FriendOfEveryone: Ok

mw-webapp/pnpm-lock.yaml
Ekaterina1994: Then these changes need to delete

mw-webapp/src/component/image/Image.module.scss
Ekaterina1994: Pease in comments write questions , suggestions, or status : done, deleted etc. Ok is not clear

mw-webapp/src/component/image/Image.tsx
Ekaterina1994: yes

mw-webapp/src/component/image/Image.tsx
Ekaterina1994: Yes? sory, I was mean
isOpen && styles.imageWrapperFullScreen

FriendOfEveryone: @Ekaterina1994 if "isOpen" will be false then className={false}

Ekaterina1994: The provided code snippet uses a conditional statement to apply a class based on the value of isOpen. Here's how it works:

className: This is a JavaScript property used to set the class names for an HTML element.
{isOpen && styles.imageWrapperFullScreen}: This is a JavaScript expression within curly braces.
isOpen: This is a boolean variable.
&&: This is the logical AND operator.
styles.imageWrapperFullScreen: This assumes styles is an object containing CSS class names. Here, it accesses the imageWrapperFullScreen class from the styles object.
If isOpen is true:

The entire expression evaluates to true because true && anything is always true.
Therefore, the class name applied will be styles.imageWrapperFullScreen.
If isOpen is false:

The expression isOpen && styles.imageWrapperFullScreen evaluates to false because false && anything is always false.
Since the entire expression evaluates to false, no additional class is applied to the element's className property. The element will only have the base classes defined outside the curly braces.
In summary:

When isOpen is false, the element won't have the styles.imageWrapperFullScreen class applied. It will retain any other classes defined for it outside the conditional statement.

FriendOfEveryone: let b = 0 && 1

FriendOfEveryone: @Ekaterina1994 let b = 0 && 1

FriendOfEveryone: @Ekaterina1994 The && operator returns the last argument if everything is true or the first value is false. In this case.
For example let b = 0 && 1. b will be 0

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: @Ekaterina1994 Without it onBlur does not work

mw-webapp/src/component/image/Image.module.scss
FriendOfEveryone: @Ekaterina1994 it is for image wrapper. Image will scale

mw-webapp/src/component/image/Image.module.scss
FriendOfEveryone: @Ekaterina1994 changed

mw-webapp/pnpm-lock.yaml
FriendOfEveryone: @Ekaterina1994 deleted

mw-webapp/src/component/image/Image.module.scss
FriendOfEveryone: @Ekaterina1994 changed

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: @Ekaterina1994 changed

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: @Ekaterina1994 changed

mw-webapp/src/component/image/Image.tsx
FriendOfEveryone: @Ekaterina1994 added close button

mw-webapp/src/component/image/Image.tsx
Ekaterina1994: According to the link https://htmlbook.ru/html/a/tabindex

Атрибут tabindex определяет последовательность перехода между ссылками при нажатии на кнопку Tab.

Do we need it?

mw-webapp/.env.example
Ekaterina1994, line 1: why did you deletes this file

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,32 @@
...
+  align-items: center;
+  justify-content: center;
+  backdrop-filter: blur(5px);
+}
+
+.closeImageButton {
+  position: fixed;
+  z-index: 99;
```
Ekaterina1994, line 15: global

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,32 @@
...
+  top: 30px;
+  width: 70px;
+  height: 70px;
+  border-radius: 50px;
+  margin: 0 auto;
```
Ekaterina1994, line 20: variables

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,32 @@
...
+  backdrop-filter: blur(5px);
+}
+
+.closeImageButton {
+  position: fixed;
+  z-index: 99;
+  top: 30px;
+  width: 70px;
+  height: 70px;
+  border-radius: 50px;
```
Ekaterina1994, line 19: We don't have such border radius by design

mw-webapp/src/component/image/Image.module.scss
```diff
@@ -1,3 +1,32 @@
...
+  margin: 0 auto;
+  background-color: #aaaaaa;
```
Ekaterina1994, line 21: use global var

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +26,41 @@ interface ImageProps {
...
+      >
+        CLOSE
+      </button>}
+      <img
+        tabIndex={0}
+        src={props.src}
+        alt={props.alt}
+        className={className}
+        data-cy={props.dataCy}
+        onClick={props.isShouldOpen ? () => setIsOpen(true) : undefined}
```
Ekaterina1994, line 59:
```suggestion
        onClick={() => props.isShouldOpen && setIsOpen(true)}
```

Issue #778, opened 2024-04-28
UX/UI_Statistics: Check-boxes in the All statistics section aren't clickable on all area because of element overlapping


evgenia7p:

evgenia7p:
Preconditions
User is authorized. User's way is created and opened. Some day reports with using tags exist

Steps
1. Open All statistics section
2. Click on the bottom area of any check-boxes in the Labels statistic blocks

Expected result
All check-boxes in the all statistics section are clickable on all area

Actual result
Check-boxes in the Labels statistic blocks of the all statistics section aren't clickable on all area because of element overlapping


![Image](https://github.com/tritonJS826/masters-way/assets/163824877/88732f4d-5d0c-4e6b-8421-f2941de4820e)

Ekaterina1994:
Need to analize libraries for charts

Ekaterina1994:
Done

evgenia7p:
Checked

PR #818, opened 2024-05-07
fix: #778 fixed clickable check-boxes in statistics section

AKiroi:
https://github.com/tritonJS826/masters-way/assets/112805668/040f7e22-1dc8-409d-878a-379d423151f1

### DIFF
--- a/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
@@ -13,5 +13,5 @@
 }
 
 .titleCheckboxIndex {
-  z-index: 1;
+  z-index: $checkboxZIndex;
 }
--- a/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
@@ -8,6 +8,7 @@
 }
 
 .titleCheckbox {
+  position: relative;
   width: 100%;
   justify-content: space-between;
 }
--- a/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
@@ -11,3 +11,7 @@
   width: 100%;
   justify-content: space-between;
 }
+
+.titleCheckboxIndex {
+  z-index: 1;
+}
--- a/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.tsx
@@ -1,4 +1,5 @@
 import {PropsWithChildren} from "react";
+import clsx from "clsx";
 import {Checkbox} from "src/component/checkbox/Checkbox";
 import {HorizontalContainer} from "src/component/horizontalContainer/HorizontalContainer";
 import {HeadingLevel, Title} from "src/component/title/Title";
@@ -29,7 +30,7 @@ interface StatisticLabelsProps {
 export const StatisticWidget = (props: PropsWithChildren<StatisticLabelsProps>) => {
   return (
     <VerticalContainer className={styles.statisticBlock}>
-      <HorizontalContainer className={styles.titleCheckbox}>
+      <HorizontalContainer className={clsx(styles.titleCheckbox, props.isEditable ? styles.titleCheckboxIndex : "")}>
         {props.title &&
         <Title
           level={HeadingLevel.h4}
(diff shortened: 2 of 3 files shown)


mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.tsx
```diff
@@ -29,7 +30,7 @@ interface StatisticLabelsProps {
 export const StatisticWidget = (props: PropsWithChildren<StatisticLabelsProps>) => {
   return (
     <VerticalContainer className={styles.statisticBlock}>
-      <HorizontalContainer className={styles.titleCheckbox}>
+      <HorizontalContainer className={clsx(styles.titleCheckbox, props.isEditable ? styles.titleCheckboxIndex : "")}>
```
tritonJS826, line 33:
```suggestion
      <HorizontalContainer className={clsx(styles.titleCheckbox, props.isEditable && styles.titleCheckboxIndex)}>
```

mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
```diff
@@ -11,3 +11,7 @@
   width: 100%;
   justify-content: space-between;
 }
+
+.titleCheckboxIndex {
+  z-index: 1;
```
tritonJS826, line 16: let's use global variables for such purpose

mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.tsx
AKiroi: I used the same logic like 
image
because props.isEditable === isCheckboxShown
image
if we have checkbox in StatisticWidget(isCheckboxShown) =>  container titleCheckbox need add z-index to up under barChartContainer with canwas. if remove this check 
image
we have titleCheckboxcontainer in all StatisticWidget, but we need it only with isCheckboxShown

AKiroi: withot this check 
image

AKiroi: suggest another way to fix this bug? let's disscus in conversation

AKiroi: its not good resolve because we have checkbox in StatisticWidget only in Modal window, if we add to another component isCheckboxShown it will be problem

mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
AKiroi: Why we need to use global variables? in global we have only 
image

AKiroi: add to glibal variable

mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.tsx
Ekaterina1994: maybe this problem will be resolved if you add z-index to checkbox,  not to wrapper titleCheckbox?
Let's try)

AKiroi: fixed
it's doesn't work(z-index) without add position:relative in titleCheckbox. With relative everything allright

Ekaterina1994 (approved):

mw-webapp/src/styles/_variables.scss
```diff
@@ -20,6 +20,7 @@ $secondaryTransitionDuration: 150ms;
 $elementHoverTransition: 500ms;
 
 // z-index
+$checkboxZIndex: 100;
```
tritonJS826, line 23: It is not truth, actually 
:)

mw-webapp/src/logic/wayPage/wayStatistics/statisticWidget/StatisticWidget.module.scss
```diff
@@ -8,6 +8,11 @@
 }
 
 .titleCheckbox {
+  position: relative;
   width: 100%;
   justify-content: space-between;
 }
+
+.titleCheckboxIndex {
+  z-index: $checkboxZIndex;
```
tritonJS826, line 17: Are you sure we don't have other options to solve the problem with clickable checkboxes?

tritonJS826, line 17: Are you sure we don't have other options to solve the problem with clickable checkboxes?

tritonJS826, line 17: Are you sure we don't have other options to solve the problem with clickable checkboxes?

mw-webapp/src/styles/_variables.scss
tritonJS826: Sent audio message to you

Issue #816, opened 2024-05-06
Add mobX to the project


Ekaterina1994:
- [x] Change Theme by using mobX as a global store
- [x] Change Language by using mobX as a global store

pp2amd6:
@tritonJS826 @Ekaterina1994 how QA Team should / can test that? :)

Ekaterina1994:
Need to check updating theme, language and user logic

EkaterinaSokolova-V:
`done

PR #826, opened 2024-05-09
refactor: #816 Move lang to store

Ekaterina1994:


mw-webapp/src/logic/wayPage/reportsTable/dayReportsTable/DayReportsTable.tsx
```diff
@@ -40,7 +40,7 @@ interface DayReportsTableProps {
  * move load logic to the parent component and share data with other components
  */
 export const DayReportsTable = (props: DayReportsTableProps) => {
```
tritonJS826, line 42: don't you need to use observer here?

Ekaterina1994: Added

tritonJS826 (approved):

PR #893, opened 2024-05-19
Mw 885/fix select width

Ekaterina1994:


mw-webapp/src/component/select/Select.module.scss
```diff
@@ -45,17 +45,15 @@ $selectPadding: 3px;
 }
 
 .SelectItem {
-  position: relative;
-  display: flex;
-  height: 25px;
-  align-items: center;
+  overflow: hidden;
+  height: 24px;
```
tritonJS826, line 49: why 24px? is it according to design?

tritonJS826 (approved):

Issue #897, opened 2024-05-20
DayReports: sort jobs, plans, problems, comments by created date on the front when creating composite way


tritonJS826:
After reloading page, order of jobs, plans, problems, comments could be changed. We shoul dhave only one flow for both of them for now

n144astya:
Environment: firefox Версия 125.0.1 (64-битная), Windows 10 Version 22H2

Actual result;
Home page: background image is not displayed


![Image](https://github.com/tritonJS826/masters-way/assets/142949669/96535519-fd5f-494a-b4c3-970242a97e8c)

Home page doesn't match the layout

![Image](https://github.com/tritonJS826/masters-way/assets/142949669/3c3f3e81-29ad-4165-8c18-9f94ac20485a)

pp2amd6:
@n144astya 

> Home page doesn't match the layout

#954

vdruda:
Done

vdruda:
Done. Migrationis done, works correctly local branch and main branch

PR #899, opened 2024-05-20
feat: #897 Add footer and update styles for about page

Ekaterina1994:


mw-webapp/src/logic/aboutProjectPage/AboutProjectPage.module.scss
```diff
@@ -1,12 +1,9 @@
...
 
-.pageWrapper {
-  max-width: $maxWidthContainer;
-  align-items: center;
-  padding: 0 $gapMedium;
-  margin: $marginMedium auto;
-  gap: $gapBig;
-}
+// Height of header. Need for add background under the header
+$mainImageMarginTop: -103px;
```
tritonJS826, line 4: better to handle it consystently for all similar pages. It will be impossible in future wheb we decided to uncrease header heigh for 5 px for example check it everywhere

Ekaterina1994: Fixed

tritonJS826 (approved):

PR #933, opened 2024-05-24
feat: #208 Create Text component

timoxa20:
Created a new Text component to render text content within a span element. This component accepts both value and placeholder props, providing dynamic rendering based on the content.


tritonJS826 (commented):
Good start) let'w wait for the Ekaterinas's review)

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Displayed values
+   */
+  value: string | number;
+
+  /**
+   * Displayed placeholder
+   */
+  placeholder: string;
```
Ekaterina1994, line 16: Text shouldn't have prop placeholder. Text can have only value

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Function to render value
+   */
+  renderValue: (value: string | number) => ReactElement;
```
Ekaterina1994, line 21: You don't need to pass function as props in this case.

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Component Text
+ */
+export const Text = (props: TextProps) => {
+  const {
+    placeholder,
+    value,
+    renderPlaceholder,
+    renderValue,
+  } = props;
```
Ekaterina1994, line 38: Delete this/ We don't use destructure for props in the project

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Function to render value
+   */
+  renderValue: (value: string | number) => ReactElement;
+
+  /**
+   * Function to render placeholder
+   */
+  renderPlaceholder: (placeholder: string) => ReactElement;
+}
```
Ekaterina1994, line 27: Need to add additional  optional prop className, because span can be differently styled in the different places

Ekaterina1994, line 27: Add prop type (for the future) - example you can find in the component button/ It will be needed for using different styled text in different places. You can create only one type right now - because design is not ready

Ekaterina1994 (changes requested):
Hey) Great! But let's add for Text component basic test and story for storybook/ You can see example in other components (files .cy.tsx, .stories.tsx)

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Params for {@link Text}
+ */
+interface TextProps {
+
+  /**
+   * Displayed values
+   */
+  value: string | number;
```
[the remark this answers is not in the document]
timoxa20: I didn't change the value because in the editable Text or editableTextaria component, ts was swearing, which could be a number. I left it like this

mw-webapp/src/component/editableText/renderSpan.tsx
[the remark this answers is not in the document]
timoxa20: I moved the logic and deleted the renderSpan file, now the Text component is responsible for the display

mw-webapp/src/component/text/Text.tsx
timoxa20: Deleted the placeholder for the Text component

mw-webapp/src/component/text/Text.tsx
timoxa20: Deleted the function

mw-webapp/src/component/text/Text.tsx
timoxa20: deleted destructure

mw-webapp/src/component/text/Text.tsx
timoxa20: Added props text to the component. To display different types of text, as well as the size of the text. Depending on what type of test is needed

mw-webapp/src/component/text/Text.cy.tsx
```diff
@@ -0,0 +1,26 @@
...
+describe("Text component", () => {
+  it("should render text when it is provided", () => {
+    cy.mount(
+      <Text
+        text={TEXT_VALUE}
+        cy={TEXT_CY}
+      />,
+    );
+    cy.get(`[data-cy=${TEXT_CY}]`).should("contain.text", TEXT_VALUE);
+  });
```
Ekaterina1994, line 15:
```suggestion
    cy.get(getDataCy(TEXT_CY)).should("contain.text", TEXT_VALUE);
  });
```

mw-webapp/src/component/text/Text.module.scss
```diff
@@ -0,0 +1,17 @@
+.Text {
```
Ekaterina1994, line 1: No need create empty styles for the future

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * Type of  text variant styles
+ */
+
+export type ThemeVariant = "clear" | "outline";
+
+/**
+ * Type of size text styles
+ */
+
+export type TextSize = "m" | "l" | "xl";
```
Ekaterina1994, line 14: We don't need it right now because we don't have design. If we will need it let's add it later

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
+import clsx from "clsx";
+import styles from "src/component/text/Text.module.scss";
+
+/**
+ * Type of  text variant styles
+ */
+
+export type ThemeVariant = "clear" | "outline";
```
Ekaterina1994, line 8: 1. let's use enum
2. Let's rename to TextType (see example in Button component) instead of ThemeVariant
3. Let's use only one value right now because design is not ready, for example "default"

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Type of size text styles
+ */
+
+export type TextSize = "m" | "l" | "xl";
+
+/**
+ * Params for {@link Text}
```
Ekaterina1994, line 17:
```suggestion
 * Text props
```

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Displayed text
+   */
+  text: string | number;
+
+  /**
+   * Displayed size text
+   */
+  size?: TextSize;
```
Ekaterina1994, line 29: Delete this props now, maybe will add in the future

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Displayed variant theme text
+   */
+  variant?: ThemeVariant;
```
Ekaterina1994, line 34: let's rename to type

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+  text: string | number;
+
+  /**
+   * Displayed size text
+   */
+  size?: TextSize;
+
+  /**
+   * Displayed variant theme text
+   */
```
Ekaterina1994, line 33:
```suggestion
  /**
   * Displayed variant theme text
   * @default {@link TextType.DEFAULT}
   */
```

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+  variant?: ThemeVariant;
+
+  /**
+   * Data attributes for cypress testing
+   */
+  cy?: string;
+}
+
+/**
+ * Component Text
```
Ekaterina1994, line 43:
```suggestion
 * Text component
```

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const Text = (props: TextProps) => {
+  return (
+    <div data-cy={props.cy}>
+      {props.text &&
```
Ekaterina1994, line 48: props.text is not optional so it's overall. Am I wrong?

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,60 @@
...
+   */
+  cy?: string;
+}
+
+/**
+ * Component Text
+ */
+export const Text = (props: TextProps) => {
+  return (
+    <div data-cy={props.cy}>
```
Ekaterina1994, line 47: Are you sure that you need so many wrappers for text? Is it possible to avoid from some of them? If you need div maybe you don't need tag <p>?

Ekaterina1994 (changes requested):

mw-webapp/src/component/text/Text.cy.tsx
timoxa20: I noticed in cypress a test for the proposed

mw-webapp/src/component/text/Text.module.scss
timoxa20: delete file text.module.cscc

mw-webapp/src/component/text/Text.tsx
timoxa20: delete TextSize

mw-webapp/src/component/text/Text.tsx
timoxa20: Replaced the type with an enum and left the default type

mw-webapp/src/component/text/Text.tsx
timoxa20: Renamed it

mw-webapp/src/component/text/Text.tsx
timoxa20: replaced by TextType

mw-webapp/src/component/text/Text.tsx
timoxa20: reversed the positions

mw-webapp/src/component/text/Text.tsx
timoxa20: obligatory text.props

mw-webapp/src/component/text/Text.tsx
timoxa20: reduced the number of wrappers by one div

mw-webapp/src/component/text/Text.cy.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  it("should not render anything when text is empty", () => {
+    cy.mount(
+      <Text
+        text=""
+        cy={TEXT_CY}
+      />,
+    );
+    cy.get(`[data-cy=${TEXT_CY}]`).should("be.empty");
```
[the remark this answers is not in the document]
timoxa20: replaced it with the proposed version

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,43 @@
+/**
+ * Enum of  text variant styles
+ */
+
+/**
+ * Displayed type theme text
+ */
```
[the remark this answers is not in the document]
timoxa20: deleted the comment

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,43 @@
...
+   * Displayed type theme text
+   * @default {@link TextType}
+   */
+  textType?: TextType.DEFAULT;
+
+  /**
+   * Data attributes for cypress testing
+   */
+  cy?: string;
+}
```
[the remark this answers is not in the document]
timoxa20: added an optional className

Ekaterina1994 (approved):

mw-webapp/src/component/text/Text.tsx
```diff
@@ -0,0 +1,47 @@
...
+   */
+  className?: string;
+}
+
+/**
+ * Text component
+ */
+export const Text = (props: TextProps) => {
+  return (
+    <div
```
Ekaterina1994, line 40: Can you please to check is it possible to use tag p or span here instead of div?

timoxa20: Replaced <div> with <p>. I also corrected the tests to render the <p> tag

Ekaterina1994 (changes requested):
Don't delete files that need in our project

Ekaterina1994 (approved):

tritonJS826:
What happened if we will have empty few spaces or empty string in the value?

mw-webapp/src/component/editableText/EditableText.tsx
```diff
@@ -150,7 +150,7 @@ export const EditableText = <T extends string | number>(props: EditableTextProps
     >
       {isEditing
         ? renderInput()
-        : renderSpan({value, placeholder: props.placeholder})
+        : <Text text={props.value ? props.value : props.placeholder} />
```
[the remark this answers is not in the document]
timoxa20: I redid the code for the proposed version

tritonJS826 (approved):

PR #934, opened 2024-05-24
style: #913 make table with border radius

algrinkevich:
PR contains 2 commits:
- Make rounded table corners;
- Refactoring: make table component styles nested according to the SCSS [Nesting rule](https://sass-lang.com/documentation/style-rules/#nesting) to prevent repeating of selectors.

### DIFF
--- a/mw-webapp/src/component/table/Table.module.scss
+++ b/mw-webapp/src/component/table/Table.module.scss
@@ -10,50 +10,50 @@ $cell-max-width: 600px;
   background-color: var(--primaryBgColor);
   border-collapse: separate;
   border-spacing: 0;
-}
 
-.table th:first-child,
-.table td:first-child {
-  border-left: $table-border;
-}
+  th {
+    padding: 5px;
+    border-top: $table-border;
+    border-right: $table-border;
+    border-bottom: $table-border;
+    background: var(--primaryBgColor);
+    background-clip: padding-box;
+    font-weight: bold;
+  }
 
-.table tr:first-child th:first-child {
-  border-top-left-radius: $primaryBorderRadius;
-}
+  td {
+    max-width: $cell-max-width;
+    padding: 8px 10px;
+    border-right: $table-border;
+    border-bottom: $table-border;
+    text-align: left;
+    vertical-align: top;
+    word-wrap: break-word;
 
-.table tr:first-child th:last-child {
-  border-top-right-radius: $primaryBorderRadius;
-}
+    pre {
+      overflow: auto;
+    }
+  }
 
-.table tr:last-child td:first-child {
-  border-bottom-left-radius: $primaryBorderRadius;
-}
+  th:first-child,
+  td:first-child {
+    border-left: $table-border;
+  }
 
-.table tr:last-child td:last-child {
-  border-bottom-right-radius: $primaryBorderRadius;
-}
+  tr:first-child th:first-child {
+    border-top-left-radius: $primaryBorderRadius;
+  }
 
-.table th {
-  padding: 5px;
-  border-top: $table-border;
-  border-right: $table-border;
-  border-bottom: $table-border;
-  background: var(--primaryBgColor);
-  background-clip: padding-box;
-  font-weight: bold;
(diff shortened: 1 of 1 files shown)


mw-webapp/src/component/table/Table.module.scss
```diff
@@ -6,29 +6,54 @@ $cell-max-width: 600px;
 .table {
   width: 100%;
   height: 100%;
-  border: $table-border;
+  border-radius: $primaryBorderRadius;
   background-color: var(--primaryBgColor);
-  border-collapse: collapse;
-}
+  border-collapse: separate;
+  border-spacing: 0;
```
Ekaterina1994, line 12: Why?

mw-webapp/src/component/table/Table.module.scss
```diff
@@ -6,29 +6,54 @@ $cell-max-width: 600px;
...
+  }
+
+  td {
+    max-width: $cell-max-width;
+    padding: 8px 10px;
+    border-right: $table-border;
+    border-bottom: $table-border;
+    text-align: left;
+    vertical-align: top;
+    word-wrap: break-word;
```
Ekaterina1994, line 32: Can you please explain how these changes connect with issue - add border-radius?

Ekaterina1994 (changes requested):

Ekaterina1994 (changes requested):
Please add screenshot to the description of the result)

mw-webapp/src/component/table/Table.module.scss
algrinkevich: It is not directly connected with "add border-radius". It just refactors code to use SCSS nested structures for reducing number of repeated selecters. It's done in a separate commit. I can revert commit if you think it's not suitable for this PR.

mw-webapp/src/component/table/Table.module.scss
```diff
@@ -6,29 +6,54 @@ $cell-max-width: 600px;
 .table {
   width: 100%;
   height: 100%;
-  border: $table-border;
```
Ekaterina1994, line 9: Can you please explain how table's border renders if you deleted this property?

mw-webapp/src/component/table/Table.module.scss
algrinkevich: border-radius doesn't work with "collapse" css property value:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius 
> "The border-radius property does not apply to table elements when border-collapse is collapse."

Since that I need to use `border-collapse: separate`, but this causes unnecessary spacing between table borders:
image

so I use "border-spacing: 0" to fix this issue.

P.S. I added the screenshot of rounded table to the PR description.

mw-webapp/src/component/table/Table.module.scss
algrinkevich: I use cell borders instead of table border. Using table borders along with "border-collapse: separate" causes issues with borders: 
- strange table corners; 
- table's border is thicker than expected.

image

mw-webapp/src/component/table/Table.module.scss
Ekaterina1994: Yea, right

mw-webapp/src/component/table/Table.module.scss
Ekaterina1994: It's okey.
But what do you think about using
border: $table-border $table-border $table-border  0;

instead of:
border-top: $table-border;
border-right: $table-border;
border-bottom: $table-border;

algrinkevich: It's impossible to do like this because border accepts [args](https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax): line-width, line-style, color.

But I found another way to make it shorter:
```css
border: $table-border;
border-left: 0;
```
Are you ok with that?

algrinkevich: Done

Issue #607, opened 2024-03-05
Move view switcher to the components (and reuse)


tritonJS826:

tritonJS826:
@Ekaterina Refactor optoin (generateViewSwitcherOptions) -> move to regular consts

Ekaterina1994:
Done

PR #963, opened 2024-05-30
feat: #607 Move view switcher to the components (and reuse)

Howl404:


mw-webapp/src/component/viewSwitcher/ViewSwitcher.cy.tsx
```diff
@@ -0,0 +1,57 @@
+import Sinon from "cypress/types/sinon";
+import {ViewOption, ViewSwitcher, ViewSwitcherProps} from "src/component/viewSwitcher/ViewSwitcher";
+import {getDataCy} from "src/utils/cyTesting/getDataCy";
+import {View} from "src/utils/LocalStorageWorker";
+import styles from "src/component/viewSwitcher/ViewSwitcher.module.scss";
+
+const VIEW_SWITCHER_CY = "view-switcher";
+const CARD_ICON_CY = "card-view";
+const TABLE_ICON_CY = "table-view";
+
```
Ekaterina1994, line 10: Let's use camelCase )

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * View types
+   */
+  view: View;
+
+  /**
+   * Tooltip's content
+   */
+  tooltipContent: TooltipProps["content"];
```
Ekaterina1994, line 22: I think it can be just string in this case. Am I wrong? Is there any  other content ?

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Icon name
+   */
+  iconName: IconProps["name"];
```
Ekaterina1994, line 27: Maybe better to use keyof typeof IconDictionary?

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
```diff
@@ -0,0 +1,95 @@
...
+      dataCy={props.dataCy}
+    >
+      {props.options.map((option) => (
+        <Tooltip
+          key={option.view}
+          position={PositionTooltip.LEFT}
+          content={option.tooltipContent}
+          dataCy={option.dataCy}
+        >
+          <button
```
Ekaterina1994, line 82: We have component Button with type icon Button. Let's try to use component instead of default button if its possible)

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Inside Tooltip props interface it can be string | ReactNode | ReactElement, I would like to have a single place of truth

mw-webapp/src/component/viewSwitcher/ViewSwitcher.cy.tsx
Howl404: Please specify, for variable names or values

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Done

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
tritonJS826: yeah. it will be convenient to use some react nodes in tooltips,

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
tritonJS826: It's a philosophical question whether viewSwitcher needs to relate on IconProps or not. On the one hand, viewSwitcher uses IconProps and logically determines the type of the field. But at the same time, we “link” these two components directly + readability decreased (2 jumps instead of one, the ability to “jump through the search” without an IDE is difficult). And most important -  it’s good to make components as independent as possible at the interface level. For example, it now seems that creating a viewSwitcher requires that the IconProps have a name field, which from a design point of view is an awkward statement that will be confusing and raise "why" questions. My personal opinion is that there is no need to inherit types if we don't have good reasons, hierarchies are really  difficult to maintain if they do not have systems. That's why some people hate OOP - with no good system the web of relation is a mess

mw-webapp/src/component/viewSwitcher/ViewSwitcher.cy.tsx
Ekaterina1994: For values in this case

Howl404: Done

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Replaced with string | ReactNode | ReactElement

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Replaced with keyof typeof IconDictionary

mw-webapp/src/component/tooltip/Tooltip.tsx
```diff
@@ -6,7 +6,7 @@ import styles from "src/component/tooltip/Tooltip.module.scss";
 /**
  * Tooltip props
  */
-interface TooltipProps {
+export interface TooltipProps {
```
Ekaterina1994, line 9:
```suggestion
interface TooltipProps {
```

mw-webapp/src/logic/allUsersPage/AllUsersPage.tsx
```diff
@@ -187,40 +184,22 @@ export const AllUsersPage = observer(() => {
...
+              tooltipContent: LanguageService.allUsers.filterBlock.cardViewTooltip[language],
+              iconName: "GridViewIcon",
+            },
+            {
+              view: View.Table,
+              tooltipContent: LanguageService.allUsers.filterBlock.tableViewTooltip[language],
+              iconName: "TableViewIcon",
+            },
+          ]}
+        />
```
Ekaterina1994, line 202: What do you think about move it to const and reuse it in all the files?

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {PositionTooltip} from "src/component/tooltip/PositionTooltip";
+import {Tooltip} from "src/component/tooltip/Tooltip";
+import {View} from "src/utils/LocalStorageWorker";
+import styles from "src/component/viewSwitcher/ViewSwitcher.module.scss";
+
+/**
+ * View option
+ */
+export interface ViewOption {
+
```
Ekaterina1994, line 15: Do we need export this interface?

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Data attribute for cypress testing
+   */
+  dataCy?: string;
+}
+
+/**
+ * ViewSwitcher props
+ */
+export interface ViewSwitcherProps {
```
Ekaterina1994, line 40: Do we need export this interface?

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Used in component test

mw-webapp/src/component/viewSwitcher/ViewSwitcher.tsx
Howl404: Used in component test

mw-webapp/src/logic/allUsersPage/AllUsersPage.tsx
Howl404: Content variable is different on each page

mw-webapp/src/component/tooltip/Tooltip.tsx
Howl404: Fixed

mw-webapp/src/logic/allUsersPage/AllUsersPage.tsx
Ekaterina1994: Well we can move OptionItems and reuse them/
For example optinCard, optionTable and etc

tritonJS826 (approved):

mw-webapp/src/logic/allUsersPage/AllUsersPage.tsx
Howl404: Done

tritonJS826 (approved):

PR #970, opened 2024-05-30
feat: #396 Add option to open image in full screen

### LINKED ISSUE #396: Make image component clickable (fullscreen image)

artiomcatruc:
This PR contains the following changes:

- Added prop isZoomable with @default: false in component Image.tsx.
- Create a if with make use of Modal component.
- Added a test for Image.tsx.
- Refactor Image.cy.tsx.
- Export const MODAL_CY.

Resolves #396


mw-webapp/src/component/image/Image.tsx
```diff
@@ -1,10 +1,12 @@
 import clsx from "clsx";
+import {Modal} from "src/component/modal/Modal";
 import styles from "src/component/image/Image.module.scss";
 
 /**
  * Props for the Image component
  */
-interface ImageProps {
+interface ImageProps
+{
```
Ekaterina1994, line 9: please return this bracket on 7 line

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,55 @@ interface ImageProps {
    * Data attribute for cypress testing
    */
   dataCy?: string;
+
+  /**
+   * Enlarge Image
+   */
+  enlargeImage?: boolean;
+
```
Ekaterina1994, line 35: We use prefix is for boolean

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,55 @@ interface ImageProps {
    * Data attribute for cypress testing
    */
   dataCy?: string;
+
+  /**
+   * Enlarge Image
+   */
+  enlargeImage?: boolean;
```
Ekaterina1994, line 34: let's rename to  isZoomable fore more readability

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,55 @@ interface ImageProps {
...
+   */
+  enlargeClassName?: string;
 }
 
 /**
  * Сomponent for displaying images
  */
 export const Image = (props: ImageProps) => {
   const className = clsx(styles.image, props.className);
+  const enlargeClassName = clsx(styles.image, props.enlargeClassName);
```
Ekaterina1994, line 48: Let's create one className
styles that can be not used you can describe as
 props.zoomable && styles.zoomable

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,55 @@ interface ImageProps {
...
+      ? (
+        <Modal
+          trigger={
+            <img
+              src={props.src}
+              alt={props.alt}
+              className={className}
+              data-cy={props.dataCy}
+            />
+          }
```
Ekaterina1994, line 60: let's move img... above return and reuse one element instead of write img 3 times)

Ekaterina1994 (changes requested):
Great!

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,55 @@ interface ImageProps {
...
+  const enlargeClassName = clsx(styles.image, props.enlargeClassName);
 
   return (
-    <img
-      src={props.src}
-      alt={props.alt}
-      className={className}
-      data-cy={props.dataCy}
-    />
+    props.enlargeImage
```
tritonJS826, line 50: 1. Let's create internal state to handle this instead of prop + external state
2. We should be able to disable zoomable behavior through props

tritonJS826 (changes requested):

mw-webapp/src/component/image/Image.tsx
artiomcatruc: done!

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Yes!

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Done!

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Done!

mw-webapp/src/component/image/Image.tsx
artiomcatruc: I can't understand this problem

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+   */
+  isDisableZoom?: boolean;
 }
 
 /**
  * Сomponent for displaying images
  */
 export const Image = (props: ImageProps) => {
-  const className = clsx(styles.image, props.className);
+  const {src, alt, dataCy, className, enlargeClassName, isZoomable, isDisableZoom} = props;
```
tritonJS826, line 51: get rid of params destructuring here, please)

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
 
-  return (
+  const [zoomable, setZoomable] = useState(isZoomable);
```
tritonJS826, line 53: * isZoomable should be changed only outside of the component.
*  we are using prefix "is" for all boolean variables
* isZoomable должно значить может ли картинка увкличивать или нет, в твоем же случае лучше использовать isZoomed

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+  useEffect(() => {
+    if (isDisableZoom) {
+      setZoomable(false);
+    } else {
+      setZoomable(isZoomable);
+    }
+  }, [isDisableZoom, isZoomable]);
+
+  const combinedClassName = clsx(styles.image, className);
+  const combinedEnlargeClassName = clsx(styles.image, enlargeClassName);
```
tritonJS826, line 64: too much classes, clsx could combine 2 and more classes with no limits this wat ```clsx(styles.1, ["styles.2"]: true.... or other ways)```, check clsx API and other components like Button

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+
+  /**
+   * Enlarge Image Styles
+   */
+  enlargeClassName?: string;
+
+  /**
+   * Enlarge Image Styles
+   */
+  isDisableZoom?: boolean;
```
tritonJS826, line 44: what the difference between isZoomable and isDisableZoom? I whink we should get rid of one of them

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+          content={enlargedImageElement}
+          className={combinedEnlargeClassName}
+        />
+      )
+      : imageElement
+  );
 };
 
+// Default props if needed
+Image.defaultProps = {
```
tritonJS826, line 98: we can handle default props pretty nice way without .defaultProps here, make isZoomable false by default

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+  const enlargedImageElement = (
     <img
-      src={props.src}
-      alt={props.alt}
-      className={className}
-      data-cy={props.dataCy}
+      src={src}
+      alt={alt}
+      className={combinedEnlargeClassName}
+      data-cy={dataCy}
```
tritonJS826, line 80: if we want to make one of props optional we don't need to programm two whole components

tritonJS826, line 80: if we want to make one of props optional we don't need to programm two whole components

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +27,75 @@ interface ImageProps {
...
+
+  /**
+   * Enlarge Image
+   */
+  isZoomable?: boolean;
+
+  /**
+   * Enlarge Image Styles
+   */
+  enlargeClassName?: string;
```
tritonJS826, line 39: i think it will be better to make all enlarge images the same for now, lets remove this prop to prohibit enlarge image view modifications outside of the component

tritonJS826 (changes requested):

mw-webapp/src/component/image/Image.tsx
tritonJS826: I mean Image component dont need to have ability to become big programmatically, let's just handle click event

tritonJS826 (changes requested):
Please, don't close conversations) We will close them usually during the review process to understand changes better)

mw-webapp/src/component/image/Image.tsx
artiomcatruc: i tried to change it here, but i'm not very sure what's wrong

mw-webapp/src/component/image/Image.tsx
artiomcatruc: 1) Yes
2) All right 
3) Changed

mw-webapp/src/component/image/Image.tsx
artiomcatruc: All right

mw-webapp/src/component/image/Image.tsx
artiomcatruc: No difference, removed isDisableZoom

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Changed to default

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Removed!

mw-webapp/src/component/image/Image.tsx
artiomcatruc: I still can't resolve out how to solve this problem

artiomcatruc: I still can't resolve out how to solve this problem

artiomcatruc:
"Please, don't close conversations) We will close them usually during the review process to understand changes better)"

Ohh,  i'm sorry. i'll be careful in the future

artiomcatruc (commented):
This is a bad commit, please see the next one

mw-webapp/src/component/image/Image.tsx
tritonJS826: use classNames

mw-webapp/src/component/image/Image.tsx
tritonJS826: described in the video below

mw-webapp/src/component/image/Image.tsx
tritonJS826: You removed isZoomable prop instead of making it work only outside)

tritonJS826:
https://github.com/tritonJS826/masters-way/assets/56690686/8c25e326-09ed-4faa-9252-04e8e22e9afe

.idea/.gitignore
Ekaterina1994, line 1: Why do you add it to repo?

.idea/inspectionProfiles/Project_Default.xml
Ekaterina1994, line 1: Are you sure that we need this file?

.idea/vcs.xml
Ekaterina1994, line 1: Delete it

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +35,66 @@ interface ImageProps {
...
+
+  /**
+   * Enlarge Image
+   */
+  isZoomable?: boolean;
+
+  /**
+   * Large Image
+   */
+  isZoomed?: boolean;
```
Ekaterina1994, line 47: Is it duplicate?

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +35,66 @@ interface ImageProps {
    * Data attribute for cypress testing
    */
   dataCy?: string;
+
+  /**
+   * Enlarge Image
```
Ekaterina1994, line 40: Describe comments in right way. You can see example in other components

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +35,66 @@ interface ImageProps {
...
-};
 
+  return (
+    props.isZoomable
+      ? (
+        <DialogRoot
+          open={props.isZoomed ?? isOpen}
+          onOpenChange={handleOpenChange}
+        >
+          <ModalTrigger>
```
Ekaterina1994, line 87: Why do you need it? We have alreaady modal component

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.tsx
artiomcatruc: <ModalTrigger>
      {imageElement}
      <ModalContent>
        
      </ModalContent>
    </ModalTrigger>
    
    Use ModalTrigger and ModalContent ? Without Modal cannot use useState

.idea/inspectionProfiles/Project_Default.xml
artiomcatruc: Removed

mw-webapp/src/component/image/Image.tsx
artiomcatruc: No, they are different, isZoomable respond to click.
isZoomed immediately enlarges the photo or you can add a button and use it with it.

mw-webapp/src/component/image/Image.tsx
artiomcatruc: Done

artiomcatruc:
I changed Modal as in the first PR. I'm using state. I can't use only className for Image, it doesn't let pnpm cypress pass through. Each prop is described. I'm waiting for your answer

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +32,72 @@ interface ImageProps {
...
-  const className = clsx(styles.image, props.className);
+  const imageClass = clsx(styles.image, props.className);
+  const [isOpen, setIsOpen] = useState(false);
+
+  /**
+   * Toggle modal state
+   */
+  const handleOpenChange = (value: boolean) => {
+    setIsOpen(value);
+  };
```
tritonJS826, line 65: useless wrapper

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +32,72 @@ interface ImageProps {
...
+
+  /**
+   * Image respond to clicks
+   */
+  isZoomable?: boolean;
+
+  /**
+   * Large Image
+   */
+  isZoomed?: boolean;
```
tritonJS826, line 44: Looks like this props required for programmatically change image size (open in modal) from outside.  We don't need it for now I think, let's remove and keep only internal state for such puprpose.

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +32,72 @@ interface ImageProps {
...
+
+  /**
+   * Control Image outside
+   */
+  onOpenChange?: (arg: boolean) => void;
```
tritonJS826, line 49: arg is too abstract naming, choose something more specific)

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +32,72 @@ interface ImageProps {
...
+        }
+        content={
+          <img
+            src={props.src}
+            alt={props.alt}
+            className={clsx(styles.zoomedImage, props.className)}
+            onClick={() => handleOpenChange(false)}
+          />
+        }
+        isOpen={props.isZoomed ?? isOpen}
```
tritonJS826, line 95: modal should work with no isOpen prop in our case

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +32,72 @@ interface ImageProps {
...
+        className={imageClass}
+        cy={props.cy}
+      />
+    )
+    : (
+      imageElement
+    );
+};
```
tritonJS826, line 103: you have 3 img with 2 cy attributes. It looks strange a little, think about it how to test it properly. I think only ine cy id required for our case.

Good. Lets add tests for new functionality and check that everything works in storybook and in the demo

.idea/.gitignore
tritonJS826, line 1: please check all files in your PR, does all of them required?

tritonJS826 (changes requested):

tritonJS826 (commented):
please add tests  to check new functionality
Why do we need Project_Default.xml in the repo? check other unnecessary files

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -51,4 +51,10 @@ describe("Image component", () => {
       .and("not.be.empty")
       .should("eq", LOGO_TEXT);
   });
+
+  it("should enlarge the image on click if isZoomable is true", () => {
+    cy.mount(createTestImage({src: logo}));
+    cy.get(getDataCy(IMAGE_CY)).click();
+    cy.get(`.${styles.image}`).should("be.visible"); // Verify the modal content
```
Ekaterina1994, line 58: Is it possible to check is modal opened or not by using modal's data-cy instead of style name

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.cy.tsx
artiomcatruc: Yes you can, I did that in a new commit, thanks

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +31,39 @@ interface ImageProps {
...
+  isZoomable?: boolean;
 }
 
 /**
- * Сomponent for displaying images
+ * Component for displaying images
  */
 export const Image = (props: ImageProps) => {
-  const className = clsx(styles.image, props.className);
+  const imageClass = clsx(styles.image, props.className);
```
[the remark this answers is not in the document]
artiomcatruc: I think this name would be better

.idea/.gitignore
artiomcatruc: All unnecessary files have been deleted and only 3 necessary ones remain

mw-webapp/src/component/image/Image.tsx
artiomcatruc: I removed this line because it is no longer needed

mw-webapp/src/component/image/Image.tsx
artiomcatruc: I removed this line because it is no longer needed

mw-webapp/src/component/image/Image.tsx
artiomcatruc: I removed this line because it is no longer needed

mw-webapp/src/component/image/Image.tsx
artiomcatruc: removed!!

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +31,41 @@ interface ImageProps {
    * Data attribute for cypress testing
    */
   dataCy?: string;
+
+  /**
+   * Image respond to clicks
+   * * @default false
```
Ekaterina1994, line 37:
```suggestion
   * @default false
```

tritonJS826:
hehe) why do you add my name to the commit?)

mw-webapp/src/component/image/Image.tsx
```diff
@@ -25,21 +31,41 @@ interface ImageProps {
...
-};
 
+  if (props.isZoomable) {
+    return (
+      <Modal
+        cy={props.cy}
+        trigger={imageElement}
+        content={imageElement}
+        className={imageClass}
+        data-cy={props.cy}
```
Ekaterina1994, line 64: What the difference between cy and data-cy props? Delete data-cy attribute

Ekaterina1994 (changes requested):

tritonJS826:
Message in TG

mw-webapp/src/component/image/Image.tsx
artiomcatruc: in here i removed data-cy because it repeats with cy

mw-webapp/src/component/image/Image.tsx
artiomcatruc: done!

Ekaterina1994 (approved):

mw-webapp/src/component/image/Image.tsx
```diff
@@ -1,11 +1,17 @@
...
 
 /**
  * Props for the Image component
  */
 interface ImageProps {
 
+  /**
+   * Data attributes for cypress testing
+   */
+  cy?: Cy;
```
tritonJS826, line 13: too much data attribute on the top level. Check component confirm.tsx for example. Keep the only ine property for testing purposes.

artiomcatruc: I left only cy at the top level, creating a CyDataImage like Confirm.tsx as an example, and also added MODAL_CY to the IMAGE_CY variable in Image.cy.tsx to use only one testing property.

mw-webapp/src/component/image/Image.tsx
```diff
@@ -1,11 +1,28 @@
...
 
+/**
+ * Data attributes for cypress testing
+ */
+interface CyDataImage extends Cy {
+
+  /**
+   * Data attribute for cypress testing
+   */
+  onZoom?: string;
```
Ekaterina1994, line 13: it's not onZoom cy, it's cy for image even if it's not zoomed
Rename to imageCy or similar

Ekaterina1994 (changes requested):

mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -1,9 +1,13 @@
 import logo from "src/assets/mastersWayLogo.svg";
 import {LOGO_TEXT} from "src/component/header/Header";
 import {Image} from "src/component/image/Image";
+import {MODAL_CY} from "src/component/modal/Modal.cy";
 import {getDataCy} from "src/utils/cyTesting/getDataCy";
 
-const IMAGE_CY = "image";
+const IMAGE_CY = {
+  dataCy: "image",
```
Ekaterina1994, line 8: Did not found dataCy prop in CyDataImage interface. Am I wrong?

Ekaterina1994: Should be something like this 

const Image_CY = {
  dataCyTrigger: "trigger",
  dataCyContent: {
    dataCyOverlay: "overlay",
    dataCyClose: "close",
    dataCyContent: "content",
    dataCyOK: "confirm",
  },
  dataCyImage: "image",
};

mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -15,40 +19,45 @@ export interface createTestImageProps {
...
     cy.mount(createTestImage({src: logo}));
-    cy.get(getDataCy(IMAGE_CY)).should("exist");
+    cy.get(getDataCy(IMAGE_CY.dataCy)).should("exist");
   });
 
   it("image component should be visible", () => {
     cy.mount(createTestImage({src: logo}));
-    cy.get(getDataCy("image")).should("be.visible");
+    cy.get(getDataCy(IMAGE_CY.dataCy)).should("be.visible");
   });
```
Ekaterina1994, line 47: 2 similar tests, lets avoid one of them?

mw-webapp/src/component/image/Image.cy.tsx
```diff
@@ -15,40 +19,45 @@ export interface createTestImageProps {
...
-    cy.get(getDataCy(IMAGE_CY))
+    cy.get(getDataCy(IMAGE_CY.dataCy))
       .invoke("attr", "alt")
       .should("exist")
       .and("not.be.empty")
       .should("eq", LOGO_TEXT);
   });
+
+  it("should enlarge the image on click if isZoomable is true", () => {
+    cy.mount(createTestImage({src: logo}));
```
Ekaterina1994, line 59: Use beforeEach(... construction to avoid repeating in each test command mount

Ekaterina1994 (changes requested):

Issue #978, opened 2024-06-01
Modals: broken modal paddings


tritonJS826:

tritonJS826:
![Image](https://github.com/tritonJS826/masters-way/assets/56690686/4ac6090c-3f33-4b3a-9dc9-9c4f905719c9)

Ekaterina1994:
Done

EkaterinaSokolova-V:
Fixed

PR #987, opened 2024-06-03
fix: #978 Modals: broken modal paddings

algrinkevich:
This PR contains fixes related to adding padding to modals on small screens (#978).

As adding padding to modals breaks statistics in modals, so I also touched this place. Now it looks like this:

### DIFF
--- a/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.module.scss
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.module.scss
@@ -6,4 +6,8 @@
   width: 100%;
   gap: $gapMedium;
   grid-template-columns: repeat(auto-fit, minmax($minWidthBlock, 1fr));
+
+  @media only screen and (max-width: $mediaWidthMobileMedium) {
+    padding: $paddingBig 4px;
+  }
 }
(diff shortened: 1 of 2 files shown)


mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.module.scss
```diff
@@ -6,4 +6,8 @@
   width: 100%;
   gap: $gapMedium;
   grid-template-columns: repeat(auto-fit, minmax($minWidthBlock, 1fr));
+
+  @media only screen and (max-width: $mediaWidthMobileMedium) {
+    padding: $paddingBig 4px;
```
Ekaterina1994, line 11: Please use only existing tokens (from global variables)

algrinkevich: Done

Ekaterina1994 (approved):

mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.module.scss
```diff
@@ -6,4 +6,8 @@
   width: 100%;
   gap: $gapMedium;
   grid-template-columns: repeat(auto-fit, minmax($minWidthBlock, 1fr));
+
+  @media only screen and (max-width: $mediaWidthMobileMedium) {
+    padding: $paddingBig $paddingSmall;
+  }
```
tritonJS826, line 12: why we added paddings here for mobile only, what about desktop?

tritonJS826: Ok)

Issue #795, opened 2024-04-30
Statistics: the chart does not show tasks completed on the first day


n144astya:

n144astya:
![Image](https://github.com/tritonJS826/masters-way/assets/142949669/ac46a874-3a91-49fc-b99c-3756b6ec1ff4)



![Image](https://github.com/tritonJS826/masters-way/assets/142949669/ea33511e-c026-4632-af8b-ec906b8c0b83)

Ekaterina1994:
Done

vdruda:
Done

PR #999, opened 2024-06-05
fix: #795  the chart does not show tasks completed on the first day

AKiroi:
fixed Statistics: the chart does not show tasks completed on the first day

### DIFF
--- a/mw-webapp/src/logic/wayPage/wayStatistics/WayActiveStatistic.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/WayActiveStatistic.tsx
@@ -94,9 +94,8 @@ export const WayActiveStatistic = observer((props: WayStatisticProps) => {
   const dayReportsReversed = [...props.dayReports].reverse();
   const startDate = dayReportsReversed[0] ? dayReportsReversed[0].createdAt : props.wayCreatedAt;
   const lastDate = props.dayReports[0] ? props.dayReports[0].createdAt : props.wayCreatedAt;
-  const startDateLastWeek = startDate <= DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate)
-    ? DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate)
-    : startDate;
+  const lastDateWeek = DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate);
+  const startDateLastWeek = startDate <= lastDateWeek ? lastDateWeek : startDate;
 
   const datesWithJobTotalTime: Map<string, number> = new Map(props.dayReports.map((report) => {
     const jobDoneTotalTime = report.jobsDone.reduce((totalTime, jobDone) => totalTime + jobDone.time, 0);
@@ -129,7 +128,7 @@ export const WayActiveStatistic = observer((props: WayStatisticProps) => {
   const averageTimeForJob = totalWayTime ? Math.round(totalWayTime / allJobs.length) : 0;
 
   const lastWeekDayReports = props.dayReports
-    .filter((dayReport) => DateUtils.roundToDate(dayReport.createdAt) > startDateLastWeek);
+    .filter((dayReport) => DateUtils.roundToDate(dayReport.createdAt) > lastDateWeek);
 
   const lastWeekJobs = lastWeekDayReports.flatMap(report => report.jobsDone);
 
--- a/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
+++ b/mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
@@ -95,11 +95,11 @@ export const WayStatistic = observer((props: WayStatisticProps) => {
   const startDate = dayReportsReversed[0] ? dayReportsReversed[0].createdAt : props.wayCreatedAt;
   const lastDate = props.dayReports[0] ? props.dayReports[0].createdAt : props.wayCreatedAt;
 
-  const lastDateMonth = DateUtils.getLastDate(AMOUNT_DAYS_IN_MONTH, lastDate);
-  const lastDateWeek = DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate);
+  const lastDateMonthReport = DateUtils.getLastDate(AMOUNT_DAYS_IN_MONTH, lastDate);
+  const lastDateWeekReport = DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate);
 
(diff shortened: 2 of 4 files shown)


mw-webapp/src/utils/DateUtils.ts
```diff
@@ -47,22 +48,28 @@ export class DateUtils {
...
-      });
-
-    return datesList;
+  public static getDatesBetween(startDate: Date, endDate: Date): string[] {
+    const dates: string[] = [];
+
+    /**
+     *GetISODate
+     */
+    const getISODate = (date: Date) => date.toISOString().split("T")[0];
```
tritonJS826, line 57: let's move it to date Date utils

mw-webapp/src/utils/DateUtils.ts
```diff
@@ -47,22 +48,28 @@ export class DateUtils {
...
+
+    const currentDate = new Date(isoStartDate);
+    const lastDate = new Date(isoEndDate);
+
+    lastDate.setDate(lastDate.getDate() + ONE_DAY);
+
+    while (currentDate < lastDate) {
+      dates.push(getISODate(currentDate));
+      currentDate.setDate(currentDate.getDate() + ONE_DAY);
+    }
```
tritonJS826, line 70: Lets  implement this logic with some array methods like forEach or map etc.? Let's avoid this kind of loops, it's will be hard to maintain

mw-webapp/src/utils/DateUtils.ts
AKiroi: fixed

mw-webapp/src/utils/DateUtils.ts
AKiroi: fixed

mw-webapp/src/utils/DateUtils.ts
```diff
@@ -48,21 +49,22 @@ export class DateUtils {
...
-         * TODO #573: Correction need because method getShortISODateValue return time in GMT format but we need UTC format
-         */
-        const correction = DAY_MILLISECONDS;
+    const isoStartDate = DateUtils.getShortISODateValue(startDate);
+    const isoEndDate = DateUtils.getShortISODateValue(endDate);
 
-        return new Date(startDate.getTime() + (i * DAY_MILLISECONDS) + correction);
-      });
+    const currentDate = new Date(isoStartDate);
+    const lastDate = new Date(isoEndDate);
```
Ekaterina1994, line 57: Can you please explain why do you need convert Date to ISO and then again convert to Date/. It's looking strange

mw-webapp/src/utils/DateUtils.ts
```diff
@@ -48,21 +49,22 @@ export class DateUtils {
...
 
-    return datesList;
+    lastDate.setDate(lastDate.getDate() + ONE_DAY);
+
+    const daysCount = (lastDate.getTime() - currentDate.getTime()) / DAY_MILLISECONDS;
+
+    const dateList = Array.from(
+      {length: daysCount},
+      (_, i) => new Date(currentDate.getTime() + (i * DAY_MILLISECONDS)));
+
```
Ekaterina1994, line 66: Too difficult to understand. Is it possible to make it easier?

Ekaterina1994 (changes requested):

mw-webapp/src/logic/wayPage/wayStatistics/WayStatistic.tsx
```diff
@@ -95,13 +95,13 @@ export const WayStatistic = observer((props: WayStatisticProps) => {
   const startDate = dayReportsReversed[0] ? dayReportsReversed[0].createdAt : props.wayCreatedAt;
   const lastDate = props.dayReports[0] ? props.dayReports[0].createdAt : props.wayCreatedAt;
 
-  const startDateLastWeek = startDate <= lastWeekDate
-    ? lastWeekDate
```
Ekaterina1994, line 99: If you don't use  const need to delete it / Or move logic to const and use const below instead of DateUtils.getLAstDate....

Ekaterina1994 (changes requested):

mw-webapp/src/utils/DateUtils.ts
AKiroi: fixed

mw-webapp/src/utils/DateUtils.ts
AKiroi: image

This is necessary because the display is in one format(ISO), while calculations are in a different format, causing data discrepancies. However, if the initial value(currentDate ) is fixed, the display will be correct.

AKiroi: the end date is fixed in the same way

AKiroi: since in the chart we only need whole dates (without minutes and seconds), when we have a fixed start and end date, we can easily calculate the number of days between dates
```
const daysCount = (lastDate.getTime() - currentDate.getTime()) / DAY_MILLISECONDS;
```

mw-webapp/src/logic/wayPage/wayStatistics/WayActiveStatistic.tsx
```diff
@@ -94,9 +94,8 @@ export const WayActiveStatistic = observer((props: WayStatisticProps) => {
   const dayReportsReversed = [...props.dayReports].reverse();
   const startDate = dayReportsReversed[0] ? dayReportsReversed[0].createdAt : props.wayCreatedAt;
   const lastDate = props.dayReports[0] ? props.dayReports[0].createdAt : props.wayCreatedAt;
-  const startDateLastWeek = startDate <= DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate)
-    ? DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate)
-    : startDate;
+  const lastDateWeek = DateUtils.getLastDate(AMOUNT_DAYS_IN_WEEK, lastDate);
```
Ekaterina1994, line 97: Could we use value of lastWeekDate - const from 49 line? Or it's 2 different values?

AKiroi: yes it's 2 different values

AKiroi: var lastWeekDate returns last week date from current date 
var lastDateWeek Reportslast week date from last date reports

AKiroi: renamed variable to lastDateMonthReport

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #993, opened 2024-06-04
Adjust pnpm monorepo


tritonJS826:
- [ ] install concurrently to the root instead using pnpm dlx
- [ ] move main scripts (for husky, starting project, etc) to the root 
- [ ] adjust caching for modules (if possible), make 3 groups of precommit scripts for husky: 1 - running when change mw-webapp module, 2 - running when changed mw-server, 3 - running always (e2e test in mw-webapp)
- [ ] you can also provide additional ideas related to improving dev experience

Ekaterina1994:
Done

PR #1010, opened 2024-06-07
feat: #993 adjust pnpm monorepo

F1zm0n:
add monorepo support, some commands and husky test run based on changed folder, modified dockerfile to work


.eslintrc
tritonJS826, line 1: why do we need eslint in the root, does it possible to move it to the webapp?

.husky/pre-commit
```diff
@@ -1,13 +1,29 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-cd mw-webapp
-pnpm lint-staged
+FRONTEND="mw-webapp/"
```
tritonJS826, line 4: Let's avoid synonims, lets be consistent with modules naming FRONTEND -> WEBAPP

.husky/pre-commit
```diff
@@ -1,13 +1,29 @@
...
+SERVER="mw-server/"
+
+if git diff --name-only --cached | grep $FRONTEND;
+  then
+    echo "FRONTEND FOLDER CHANGED --> starting husky..."
+    cd mw-webapp
+    pnpm lint-staged
+    pnpm dlx concurrently \
```
tritonJS826, line 12: we can intall concurrently and  in this case we don't need to load each evety time (get rid of dlx)

.husky/pre-commit
```diff
@@ -1,13 +1,29 @@
...
+        "pnpm type-check" \
+        "pnpm test" \
+        "pnpm build-storybook" \
```
tritonJS826, line 15: let's move such scripts to the root level and let's avoid to use "cd" command in such scripts, I hope it will be convenient

.husky/pre-commit
```diff
@@ -1,13 +1,29 @@
...
 
 pnpm dlx concurrently \
-    "pnpm type-check" \
-    "pnpm cypress:e2e:run && pnpm cypress:component:run" \
-    "pnpm test" \
-    "pnpm build-storybook" \
-    "pnpm build" \
-    # "cd ../mw-server && go build main.go"
\ No newline at end of file
+        "pnpm cypress:e2e:run && pnpm cypress:component:run"
```
tritonJS826, line 27: we don't need to run components tests if just server changed

mw-webapp/package.json
tritonJS826, line 1: why do you remove dependencies from this module? I think it should be declared on the webapp module level, not on the root

mw-webapp/Dockerfile
```diff
@@ -1,8 +1,13 @@
...
-WORKDIR /mw-webapp
-COPY . .
+WORKDIR /masters-way
+COPY ./package.json .
+COPY ./pnpm-workspace.yaml .
+COPY ./pnpm-lock.yaml .
+COPY ./.eslintrc .
 RUN npm i -g pnpm@8.15.2
 RUN pnpm install --frozen-lockfile
+COPY ./mw-webapp ./mw-webapp
```
tritonJS826, line 9: it's strange a little that we need to copy global package.json for running webapp module. I think it is required because you removed dependencies from the webapp level. Let's move it back - they should be declared on the webapp level but installed in the rootby monorepo

package.json
```diff
@@ -0,0 +1,93 @@
+{
+  "name": "masters-way",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
```
tritonJS826, line 5: why do we need main: index,js here? looks like mistake

package.json
```diff
@@ -0,0 +1,93 @@
...
+    "@radix-ui/react-avatar": "^1.0.4",
+    "@radix-ui/react-dialog": "1.0.5",
+    "@radix-ui/react-icons": "1.3.0",
+    "@radix-ui/react-progress": "^1.0.3",
+    "@radix-ui/react-select": "^2.0.0",
+    "@radix-ui/themes": "2.0.3",
+    "@tanstack/react-table": "8.12.0",
+    "@testing-library/jest-dom": "6.4.2",
+    "@testing-library/react": "14.2.1",
+    "@testing-library/user-event": "14.5.2",
```
tritonJS826, line 24: it does not look like global dependencies, be careful with them, I think it will be better to move specific dependencies back to the webapp

pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
tritonJS826, line 1: we also have mw-server module in the root

tritonJS826 (changes requested):

.eslintrc
F1zm0n: docker build is failing for some reason

package.json
```diff
@@ -0,0 +1,87 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "start:web": "pnpm --filter 'mw-webapp' run start",
+    "type-check:web": "pnpm --filter 'mw-webapp' type-check",
+    "test:web": "pnpm --filter 'mw-webapp' test",
+    "build-storybook:web": "pnpm --filter 'mw-webapp' build-storybook",
+    "build:web": "pnpm --filter 'mw-webapp' build",
+    "cypress:component:run:web": "pnpm --filter 'mw-webapp' cypress:component:run",
+    "cypress:e2e:run:web": "pnpm --filter 'mw-webapp' cypress:e2e:run"
```
tritonJS826, line 12: yeah, good idea, I can suggest to name it next way - `{moduleName}:{command}:{commandModeIfRequired}`

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
...
+  - mw-webapp
+  - masters-way
```
tritonJS826, line 3: somethisng wrong here, we don't have package masters-way as I remember, but we have mw-server instead

mw-webapp/src/FirebaseAnalytics.tsx
```diff
@@ -9,6 +9,7 @@ export const FirebaseAnalytics = () => {
   const location = useLocation();
 
   useEffect(() => {
+
```
tritonJS826, line 12: why do we need this empty line?

mw-webapp/package.json
```diff
@@ -132,5 +99,38 @@
     "vite": "5.2.11",
     "vite-plugin-eslint": "1.8.1",
     "vite-tsconfig-paths": "4.3.2"
+  },
+  "engines": {
```
tritonJS826, line 103: why did you move all this stuff below dependencies?)

mw-webapp/package.json
```diff
@@ -132,5 +99,38 @@
...
+    "node": "20.11.0",
+    "pnpm": "8.15.2"
+  },
```
tritonJS826, line 106: I see the same engines requirements (and packageManager, check for other) in the root package.json, do we need duplication?

pnpm-workspace.yaml
F1zm0n: I can remove masters-way module, i added the root module but i didn't add server because it doesn't have package.json yet

mw-webapp/package.json
F1zm0n: I needed it to build app in docker containers, otherwise build would be too complex

mw-webapp/package.json
F1zm0n: done

mw-webapp/src/FirebaseAnalytics.tsx
F1zm0n: done

pnpm-workspace.yaml
F1zm0n: done

package.json
F1zm0n: done

mw-webapp/package.json
tritonJS826: no,  I just see that you moved lines from 27-59 to 103-134

mw-webapp/src/FirebaseAnalytics.tsx
tritonJS826: dont see changes

pnpm-workspace.yaml
tritonJS826: does all modules shuold have package.json?

tritonJS826 (approved):

Issue #997, opened 2024-06-05
Add limits to the database


tritonJS826:
https://docs.google.com/spreadsheets/d/1rtwQJgdxyqh2uDWpNzCrxiUORe49oxISsd2shuq0UMU/edit#gid=0

PR #1016, opened 2024-06-08
fix: #997 Add limits to the database

Moon1it:
Created migration files. Set character limits. Values that exceeded the limits were truncated according to the limits.


tritonJS826:
limits to fields added

Issue #945, opened 2024-05-26
Refactor tooltip component: use radix or refactor custom


Ekaterina1994:

Ekaterina1994:
Done

vdruda:
Done

PR #1112, opened 2024-06-18
fix: #945 Refactored tooltip, used radix

Ekaterina1994:


mw-webapp/src/component/tooltip/Tooltip.module.scss
```diff
@@ -1,74 +1,70 @@
...
-  transform: translateY(0) translateX(-50%);
-}
-
-.bottomRight {
-  transform: translateY(0) translateX(50%);
-}
-
-.topLeft {
-  transform: translateY(calc(-100% - $marginBig)) translateX(-50%);
+@keyframes slideUpAndFade {
```
tritonJS826, line 36: !

Ekaterina1994: Done

PR #1117, opened 2024-06-19
feat: #1007 paralleled requests and data conversion

F1zm0n:
tried to parallel the interactions and mapping didn't implement test or benches, just wanted to make sure i am moving in a right direction

### DIFF
--- a/mw-server/services/user.service.go
+++ b/mw-server/services/user.service.go
@@ -203,7 +203,6 @@ func dbFavoriteWaysToDbWays(rawWay []dbb.GetFavoriteWaysByUserIdRow) []dbWay {
 		}
 	})
 }
-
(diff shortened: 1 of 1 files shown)


Moon1it:
In order for parallelization to work correctly, a connection pool to the database is needed. Currently, there is only one connection to the database. Therefore, each goroutine will wait for the database connection to be available.

Additionally, if you use goroutines, you need to synchronize their execution to handle database errors properly and avoid deadlocks.

I think this task can be accomplished using only SQL. We can try to minimize the number of queries.

tritonJS826:
let's test it with load test locally with different settings (parallel 1, 5, 10, 20 requests, 10, 50 , 100, 300 total etc). Now we have connction pool so parallelization could increase speed

Issue #1057, opened 2024-06-13
Align design for components


Ekaterina1994:
Check and update design of all components (check tokens)

- [x] Button (@algrinkevich)
- [x] Checkbox (@AKiroi)
- [x] InProgress (@AKiroi)
- [x] Input (@AKiroi)
- [x] Tooltip (@AKiroi, @Ekaterina1994 )
- [x] Toggle (@AKiroi)
- [x] Tags (@algrinkevich)
- [x] Select (@AKiroi)
- [x] Links (@algrinkevich)
- [x] Dropdown (@AKiroi)
- [x] Accoerdion (@algrinkevich)
- [x] Avatar  (@AKiroi)
- [x] Footer (@algrinkevich )
- [x] Header (@AKiroi)
- [ ]

Ekaterina1994:
Done

n144astya:
done

PR #1121, opened 2024-06-19
feat: #1057 align design for Link component

algrinkevich:
Aligned styles with the design for Link component.

### DIFF
--- a/mw-webapp/src/component/userCard/UserCard.module.scss
+++ b/mw-webapp/src/component/userCard/UserCard.module.scss
@@ -90,5 +90,8 @@ $userCardAdditionalHeight: 90px;
 
 .cardLink {
   font-size: $fontSizeBig;
-  text-decoration: none;
+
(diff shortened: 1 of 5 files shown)


mw-webapp/src/component/userCard/UserCard.module.scss
```diff
@@ -90,5 +90,8 @@ $userCardAdditionalHeight: 90px;
 
 .cardLink {
   font-size: $fontSizeBig;
-  text-decoration: none;
```
Ekaterina1994, line 93: If you delete this how will be look user card with hover? All text inside will be underlined?

algrinkevich: It won't be underlined, the 95th row below prevents from this behavior

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #1095, opened 2024-06-17
Punctuation and spelling errors in the tooltip of the "Status" column


SayaKik:
## Description

<Punctuation error in a sentence and an extra letter in the word "завершен">

## Preconditions

< 1. At least one user and some mentoring ways of this user exist.                                                                                                                                                                                  
   2. The status filter (select) sets as "All".                                                                                                                                                                                                
   3. The user has created at least one collection of ways.           >

## Steps to reproduce

<1. While on the website of the "Masters-way" application, Click Navigation burger menu and select All users item.                                                                                                                            
  2. Open user page (Personal area).                                                                                                                                                                                                                                                     
  3. Select the required collection ways.                                                                                                                                                                                         
  4. Under the selected collection, click the button to switch to table mode (table icon) and Hover over the  'Status'  title of the way table.       >

## Actual result

< When you hover your cursor over the path table header "Status" a tooltip will appear "Путь считается заброшенным, если он не завершен, но отчеты не заводились более 14 дней". >

## Expected result

< When you hover your cursor over the path table header "Status" a tooltip will appear "Путь считается заброшенным (отсутствие запятой) если он не заверНшен, но отчеты не заводились более 14 дней".>

## 

![Image](https://github.com/tritonJS826/masters-way/assets/169391750/1742d853-3e56-402e-84a2-e6a30bc61b12)



< add any other context about the problem here >

Ekaterina1994:
Done

evgenia7p:
There is still no comma after the word 'заброшенным'

Ekaterina1994:
Fixed, sorry

evgenia7p:
Checked

PR #1144, opened 2024-06-20
fix: #1095 Description error tooltip text

timoxa20:
Fix Punctuation error in a sentence and an extra letter in the word


mw-webapp/src/dictionary/AllWaysPageContent.json
```diff
@@ -104,7 +104,7 @@
         "ka": "ბოლო ანგარიშის შექმნის თარირი"
       },
       "status": {
-        "ru": "Путь считается заброшенным если он не заверншен, но отчеты не заводились более 14 дней",
+        "ru": "Путь считается заброшенным если он не завершен, но отчеты не заводились более 14 дней",
         "en": "The path is abandoned if it is not completed, but has not been edited in the last 14 days",
```
Ekaterina1994, line 108:
```suggestion
        "en": "The way is abandoned if it is not completed, but has not been edited in the last 14 days",
```

Ekaterina1994 (changes requested):

mw-webapp/src/dictionary/AllWaysPageContent.json
Ekaterina1994: Need comment

timoxa20: It has been fixed.Here?

Ekaterina1994 (approved):

PR #1152, opened 2024-06-21
feat: #1057 align design for footer

algrinkevich:
Aligned styles with the design for Footer component.

P.S. Perhaps the social media icons should be moved to a separate component, because they are at least used in the Sidebar.


mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,36 +1,44 @@
 @import "src/styles/variables";
 
-$userNameFontSize: 16px;
+$logoHeight: 55px;
+$privacyPaddingTop: 10px;
```
Ekaterina1994, line 4: Let's move padding = 10px to global var and to tokens in figma.

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,36 +1,44 @@
...
+$privacyPaddingBottom: 12px;
```
Ekaterina1994, line 5: Is it possible to use 10px? Please ask designers

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,36 +1,44 @@
...
+$iconSize: 32px;
+$privacyLineHeight: 140%;
 
 .footer {
   background-color: var(--quaternaryBgColor);
+  margin-inline: $paddingMedium;
```
Ekaterina1994, line 11: 1. Why do we need margin left and right for footer? If we need then lets use common margin instead of margin-inline.
2.Margin is not padding, so var is wrong

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,36 +1,44 @@
...
 
 .footerWrapper {
   width: 100%;
   max-width: $maxWidthContainer;
+  align-items: flex-start;
   justify-content: space-between;
-  padding: $paddingBig $paddingMedium;
   border-bottom: $primaryBorder solid var(--secondaryStrokeColor);
   margin: 0 auto;
+  padding-block: $paddingBig;
```
Ekaterina1994, line 21: I understand that  padding-block and margin-inline are more modern properties and that set values depends on text oriented.  But if we will use this properties we should rewrite all styles for these properties/ Otherwise some block could be broken.  What do you think?

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,36 +1,44 @@
...
-  padding: $paddingBig $paddingMedium;
   margin: 0 auto;
   gap: $gapLarge;
+  padding-block: $privacyPaddingTop $privacyPaddingBottom;
 }
 
 .privacyLink {
   color: var(--secondaryTextColor);
   font-size: $fontSizeBig;
+  line-height: $privacyLineHeight;
```
Ekaterina1994, line 35: All one line texts we use without line-heigth (auto) Discussed about this with designers.

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -39,14 +47,22 @@ $userNameFontSize: 16px;
   gap: $gapSmall;
 }
 
-.socialMediaIcons {
+.socialMediaText {
+  font-weight: 500;
```
Ekaterina1994, line 51: var

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -39,14 +47,22 @@ $userNameFontSize: 16px;
...
 }
 
 .socialMediaIcon {
-  border: 2px solid var(--secondaryStrokeColor);
-  border-radius: $tertiaryBorderRadius;
+  display: block;
   background-color: var(--iconBgColor);
-  fill: var(--quaternaryBgColor);
-  stroke: none;
+  stroke: var(--quaternaryBgColor);
```
Ekaterina1994, line 67: stroke is not Background

mw-webapp/src/component/footer/Footer.tsx
```diff
@@ -54,27 +51,44 @@ export const Footer = (props: FooterProps) => {
             level={HeadingLevel.h4}
             text={LanguageService.home.socialMedia[props.language]}
             placeholder=""
+            classNameHeading={styles.socialMediaText}
```
Ekaterina1994, line 54: title by default has weight 500? whe need to additional classname?

mw-webapp/src/component/footer/Footer.tsx
```diff
@@ -54,27 +51,44 @@ export const Footer = (props: FooterProps) => {
...
             <Link
               path="https://linkedin.com/company/masters-way-project"
               isNewTab
             >
-              <Icon
-                size={IconSize.MEDIUM}
-                name="LinkedinIcon"
-                className={styles.socialMediaIcon}
-              />
+              <div className={styles.iconWrapper}>
```
Ekaterina1994, line 61: why need additional wrapper?

mw-webapp/src/dictionary/CommonContent.json
```diff
@@ -23,6 +23,12 @@
     "ua": "Політика конфіденційності",
     "ka": "Კონფიდენციალურობის პოლიტიკა"
   },
+  "termsAndConditions": {
+    "ru": "Условия и положения",
```
Ekaterina1994, line 27: We don;t have this right now? let's don't create text that we can't use)

Ekaterina1994 (changes requested):

mw-webapp/src/assets/icons/YoutubeIcon.tsx
```diff
@@ -9,6 +9,9 @@ export const YoutubeIcon = (props: IconProps) => {
       xmlns="http://www.w3.org/2000/svg"
       viewBox="0 0 24 24"
       fill="none"
+      strokeWidth="2"
+      strokeLinecap="round"
+      strokeLinejoin="round"
       className={props.className}
     >
       <path d="M22.54 6.42a2.78 2.78 0 0 0-1.94-2C18.88 4 12 4 12 4s-6.88 0-8.6.46a2.78 2.78 0 0 0-1.94 2A29 29
```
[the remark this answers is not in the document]
algrinkevich: How to use CSS var() in .tsx file?

I would remove this `fill` field from this .tsx file, as it doesn't affect anything, because a polygon without `fill` field is transparent by default and the parent (svg) has desirable color. What do you think about it?

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Waiting for a response from the designer.

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Using this margin I solve one problem: when the screen's width is reduced (less than 1560px), the logo and social media block with icons are glued to the sides of the screen. I added them space to "breathe".


Actually, I found the bug here, I will try to use here another approach.

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: This line is related to the bug (I mention in [other](https://github.com/tritonJS826/masters-way/pull/1152#discussion_r1651456526) reply), so I will try to rewrite it without padding-block.

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: For stroke we don't have a variable that will be dark and won't be changes in dark/light mode. I'm waiting for a response from the designers

mw-webapp/src/component/footer/Footer.tsx
algrinkevich: By default it's 700, but I need 500 in this place (as in the design).

mw-webapp/src/component/footer/Footer.tsx
algrinkevich: If we use border-radius on a real icon (as we have now on the prod), the icon is cropped, so I added a wrapper which creates a circle and have space between the actual icon and the circle's border.

mw-webapp/src/dictionary/CommonContent.json
algrinkevich: No, we don't have it. Ok :( I thought it would be needed in the future

mw-webapp/src/component/footer/Footer.tsx
algrinkevich: Also:
> P.S. Perhaps the social media icons should be moved to a separate component, because they are at least used in the Sidebar.

mw-webapp/src/component/footer/Footer.module.scss
Ekaterina1994: Не смогла повторить , вроде ничего не прижимается на демо. Давай избегать любых марджинов, если нужно, то лучше использовать паддинги
Screenshot 2024-06-24 22:22:39

mw-webapp/src/component/footer/Footer.tsx
Ekaterina1994: Можно вынести. @tritonJS826 как лучше?

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Прижимается при моей реализации)
В демо отсуп есть (не прижимается), но линейка не выровнена с логотипом и "Политикой конфиденциальности", как по дизайну

mw-webapp/src/assets/icons/YoutubeIcon.tsx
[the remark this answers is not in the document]
algrinkevich: Удалила свойство

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Designer decided to apply 8px for padding top and bottom

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Designer decided to apply 8px for padding top and bottom

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: I fixed the bug, 'inline' and 'block' are gone

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: 'inline' and 'block' are gone

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Done

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Changed to a new variable var(--tertiaryIconColor), which also added to global variables

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,52 +1,72 @@
 @import "src/styles/variables";
 
-$userNameFontSize: 16px;
+$logoHeight: 55px;
+$iconSize: 32px;
+$privacyLineHeight: 140%;
+$socialTextFontWeight: 500;
```
Ekaterina1994, line 6: $fontWeightSemibold

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,52 +1,72 @@
 @import "src/styles/variables";
 
-$userNameFontSize: 16px;
+$logoHeight: 55px;
+$iconSize: 32px;
+$privacyLineHeight: 140%;
```
Ekaterina1994, line 5: global $lineHeightPrimary

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,52 +1,72 @@
...
   background-color: var(--quaternaryBgColor);
 }
 
 .footerWrapper {
-  width: 100%;
   max-width: $maxWidthContainer;
-  justify-content: space-between;
-  padding: $paddingBig $paddingMedium;
-  border-bottom: $primaryBorder solid var(--secondaryStrokeColor);
+  box-sizing: content-box;
```
Ekaterina1994, line 15: let's don't use content-box. It's not consisnently that in all project we use border-box and here content box

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,52 +1,72 @@
...
+  flex-direction: column;
+  padding: 0 $paddingMedium;
   margin: 0 auto;
+  gap: 0;
```
Ekaterina1994, line 19: if gap 0 then you can delete property

mw-webapp/src/component/footer/Footer.module.scss
```diff
@@ -1,52 +1,72 @@
...
+.socialLinks {
+  align-items: flex-start;
+  justify-content: space-between;
+  padding: $paddingBig 0;
+}
+
+.ruler {
+  height: $rulerHeight;
+  border: none;
+  background-color: var(--secondaryStrokeColor);
```
Ekaterina1994, line 31: if it's background then it's not Stroke

mw-webapp/src/component/footer/Footer.module.scss
Ekaterina1994: Delete please line height for one line texts

```suggestion
```

mw-webapp/src/component/footer/Footer.module.scss
Ekaterina1994: Is it possible to use border-top or border-bottom instead of background color?

Ekaterina1994 (changes requested):

mw-webapp/src/component/footer/Footer.tsx
tritonJS826: Is you always need a wrapper to use a component - it is not the best practice. So as I understand question - if it is common stiles - we can move it to separate component

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Done

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Done

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: Ok, I removed it. But now I need to use calc() function for max-width to have the same behavior. As for me, the code was easier to read with "box-sizing: content-box;"

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: I can't because horizontalContainer has gap:8px by default, and I don't need this value

mw-webapp/src/component/footer/Footer.module.scss
algrinkevich: I did as you said, but I like previous version better. Because when you read a code, you see that there is a ruler, it has 1px of height, it has a color, no borders. And semantically ruler is a stroke, I don't see any problems to use background-color for it

Ekaterina1994 (approved):

tritonJS826 (approved):

PR #1153, opened 2024-06-21
fix: #1057 add align design header

AKiroi:
align design header


mw-webapp/src/component/header/Header.module.scss
```diff
@@ -1,6 +1,11 @@
 @import "src/styles/variables";
 
-$userNameFontSize: 16px;
+$gapHeaderExtraLarge: 100px;
```
Ekaterina1994, line 3: Is it unique gap? Please check/ If yes then lets discuss with designers do we really need it) f not then need create global var

mw-webapp/src/component/header/Header.module.scss
```diff
@@ -13,38 +18,36 @@ $userNameFontSize: 16px;
...
 }
 
 .burgerMenu {
-  width: 24px;
-  height: 24px;
+  width: $burgerMenuSize;
+  height: $burgerMenuSize;
   margin-top: 2px;
   cursor: pointer;
   fill: var(--primaryTextColor);
```
Ekaterina1994, line 53: fill is not text

mw-webapp/src/component/header/Header.module.scss
```diff
@@ -75,7 +78,7 @@ $userNameFontSize: 16px;
 }
 
 .socialMediaIcon {
-  border: 2px solid var(--secondaryStrokeColor);
+  border: $secondaryBorder solid var(--secondaryStrokeColor);
   border-radius: $tertiaryBorderRadius;
   background-color: var(--iconBgColor);
   fill: var(--quaternaryBgColor);
```
Ekaterina1994, line 84: fill is not bg

mw-webapp/src/component/header/Header.tsx
```diff
@@ -195,25 +195,24 @@ export const Header = (props: HeaderProps) => {
...
-          value={props.language}
-          name="language"
-          options={languageOptions}
-          onChange={props.setLanguage}
-          className={styles.selectLanguage}
-        />
-
-        <HorizontalContainer className={styles.rightBlock}>
+      <HorizontalContainer className={styles.headerButtonsContainer}>
+        <HorizontalContainer className={styles.headerLeftBlock}>
```
Ekaterina1994, line 199: let's rename for more semantic (without lefy right, etc) Because if we will replace these element then classes will b e wrong

mw-webapp/src/styles/_variables.scss
```diff
@@ -24,6 +20,7 @@ $tagBorderRadius: 1000px;
 
 // Stroke, border
 $primaryBorder: 1px;
+$secondaryBorder: 2px;
```
Ekaterina1994, line 23: Where do we need it? Designers told that everywhere should be stroke should be 1px

Ekaterina1994 (changes requested):

mw-webapp/src/component/header/Header.module.scss
AKiroi: fixed

mw-webapp/src/component/header/Header.module.scss
AKiroi: "fill" is an attribute or property that defines the color used to paint the interior of a shape. you mean we need to create new token quanternaryIconColor like quanternaryBgColor?

mw-webapp/src/styles/_variables.scss
AKiroi: fixed

mw-webapp/src/component/header/Header.tsx
AKiroi: fixed

mw-webapp/src/component/header/Header.module.scss
AKiroi: waiting for a response from the designer

mw-webapp/src/component/header/Header.module.scss
Ekaterina1994: no, all okey, my mistake

mw-webapp/src/component/header/Header.module.scss
AKiroi: after response add new global variable $gapExtraExtraLarge: 96px;

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #1099, opened 2024-06-17
User personal data: Cursor is displayed incorrectly


evgenia7p:
Preconditions
User personal area page is opened

Steps
1.	Hover over user name
2.	Hover over e-mail
3.	Hover over ‘Mentor’ word
4.	Hover over ‘No tags’ phrase
5.	Hover over 'About' word
6.	Hover over 'About' line

Expected result
The text cursor must be enabled only in the user name and 'About' lines

Actual result
The text cursor is enabled in the user name, e-mail, ‘Mentor’ word, ‘No tags’ phrase, 'About' word and 'About' line

https://github.com/tritonJS826/masters-way/assets/163824877/793526a0-c75c-42df-b324-2064a4976c26

Ekaterina1994:
Deprecated. It's not a bug

vdruda:
Done. Not a bug, remove to Done.

PR #1181, opened 2024-06-23
fix: #1099 User personal data: Cursor is displayed incorrectly

timoxa20:
Fix Changed the text cursor for the user block, added a cursor in the user name

### DIFF
--- a/mw-webapp/src/logic/userPage/UserPage.module.scss
+++ b/mw-webapp/src/logic/userPage/UserPage.module.scss
@@ -36,6 +36,7 @@
 .userAboutBlock {
   flex-wrap: wrap;
   gap: $gapBig;
+  cursor: default;
 }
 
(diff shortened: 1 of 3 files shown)


mw-webapp/src/logic/userPage/UserPage.module.scss
Ekaterina1994, line 1: Please check all fields in the User info section. For example email is not editable but cursor is changing

Ekaterina1994 (changes requested):

mw-webapp/src/logic/userPage/UserPage.module.scss
timoxa20: Fix Changed the text cursor for the body, changed cursor where it is possible to edit fields

Ekaterina1994:
Deprecated

Issue #1174, opened 2024-06-22
Add base routing to the app


tritonJS826:
- [x] all ways
- [x] personal area
- [x] all users
- [x] home
- [x] settings
- [x]  about project

tritonJS826:
Done

tritonJS826:
Done

PR #1182, opened 2024-06-23
feat: #1174 Add base routing to the app

marluk1:


mw-mobile/lib/main.dart
```diff
@@ -1,25 +1,20 @@
...
 }
 
 class MyApp extends StatelessWidget {
-  const MyApp({super.key});
-
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
-      title: 'Flutter Neumorphism Demo',
+      title: 'E-learning App',
```
tritonJS826, line 12: Masters way?

mw-mobile/lib/routes/routes.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:flutter/material.dart';
+import '../screens/welcome_screen.dart';
+import '../screens/all_ways_screen.dart';
+import '../screens/personal_area_screen.dart';
+import '../screens/all_users_screen.dart';
+import '../screens/home_screen.dart';
+import '../screens/settings_screen.dart';
+import '../screens/about_project_screen.dart';
```
tritonJS826, line 8: Is it standard approach to use relative path in imports for dart? It's more convenient usually to use absolute path (as in typescript, java etc.) what do you think?

mw-mobile/lib/routes/routes.dart
```diff
@@ -0,0 +1,30 @@
...
+
+class AppRoutes {
+  static const String welcome = '/welcome';
+  static const String allWays = '/all_ways';
```
tritonJS826, line 12: Is it conventional for dart to use snake_case for naming?  It will be nice to use the same names as on the web - with camelCase. What do you think?

Issue #1006, opened 2024-06-06
Optimise back-db communication (way.service.GetPopulatedWayById, way.service.GetPlainWayById)


tritonJS826:
look like the are too many requests to the database, lets decrease it (only 1 will be great if possible)
you can play with benchmarks, it will be nice to see this performance optimization with some numbers
before refactoring it will be nice to add a test

tritonJS826:
not necessary for now

PR #1184, opened 2024-06-24
fix: #1006 Refactored GetPlainWayById to remove subqueries and aggreg…

yokaracho:
…ate data in a single query for improved performance.

### DIFF
--- a/mw-server/services/way.service.go
+++ b/mw-server/services/way.service.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	dbb "mwserver/db/sqlc"
 	dbbPGX "mwserver/db_pgx/sqlc"
 	"mwserver/schemas"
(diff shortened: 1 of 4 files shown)


mw-server/db/query/ways.query.sql
```diff
@@ -240,3 +240,47 @@ RETURNING *,
 -- name: DeleteWay :exec
 DELETE FROM ways
 WHERE uuid = $1;
+
+-- name: GetWayDetailsById :many
+SELECT
+    w.uuid AS way_uuid,
```
tritonJS826, line 246: let's be convenient and use full table names

mw-server/services/way.service.go
```diff
@@ -291,51 +292,72 @@ func UpdateWayIsCompletedStatus(db *dbb.Queries, ctx context.Context, wayUuid uu
...
-	mentors := lo.Map(mentorsRaw, func(dbMentor dbb.User, i int) schemas.UserPlainResponse {
-		return schemas.UserPlainResponse{
-			Uuid:        dbMentor.Uuid.String(),
-			Name:        dbMentor.Name,
-			Email:       dbMentor.Email,
-			Description: dbMentor.Description,
-			CreatedAt:   dbMentor.CreatedAt.Format(util.DEFAULT_STRING_LAYOUT),
-			ImageUrl:    dbMentor.ImageUrl,
-			IsMentor:    dbMentor.IsMentor,
+	way := wayDetails[0]
```
tritonJS826, line 304: why it is a list?

mw-server/db/query/ways.query.sql
```diff
@@ -240,3 +240,47 @@ RETURNING *,
 -- name: DeleteWay :exec
 DELETE FROM ways
 WHERE uuid = $1;
+
+-- name: GetWayDetailsById :many
```
tritonJS826, line 244: why :many

tritonJS826, line 244: why do we need another one method? Why can't we just improve old one?

mw-server/db/query/ways.query.sql
```diff
@@ -240,3 +240,47 @@ RETURNING *,
...
+    w.created_at AS way_created_at,
+    w.estimation_time AS way_estimation_time,
+    w.is_completed AS way_is_completed,
+    w.copied_from_way_uuid AS way_copied_from_way_uuid,
+    w.is_private AS way_is_private,
+    ARRAY(
+            SELECT composite_ways.child_uuid
+            FROM composite_ways
+            WHERE composite_ways.parent_uuid = w.uuid
+    )::VARCHAR[] AS children_uuids,
```
tritonJS826, line 259: After last updateds we dont need children_uuids anymore, let's use childrenAmount

tritonJS826 (changes requested):

tritonJS826:
lets use lo lib for interations

tritonJS826:
lets add a little for at least one endpoints uses this method

mw-server/db/query/ways.query.sql
yokaracho: done

mw-server/db/query/ways.query.sql
yokaracho: done

yokaracho: done

mw-server/db/query/ways.query.sql
yokaracho: done

mw-server/services/way.service.go
```diff
@@ -303,60 +304,80 @@ func UpdateWayIsCompletedStatus(db *dbb.Queries, ctx context.Context, wayUuid uu
...
+	mentorsRaw, ok := wayDetails.Mentors.([]interface{})
+	if !ok {
+		return schemas.WayPlainResponse{}, fmt.Errorf("failed to assert type for mentors")
+	}
 
-	copiedFromWayUuid := util.MarshalNullUuid(way.CopiedFromWayUuid)
-	mentorsRaw, _ := db.GetMentorUsersByWayId(ctx, way.Uuid)
+	mentors := lo.Map(mentorsRaw, func(mentor interface{}, i int) schemas.UserPlainResponse {
+		mentorMap := mentor.(map[string]interface{})
+		createdAt, err := time.Parse("2006-01-02T15:04:05", mentorMap["created_at"].(string))
```
tritonJS826, line 323: let's save all available formats somewhere nearby

mw-server/services/way.service.go
```diff
@@ -303,60 +304,80 @@ func UpdateWayIsCompletedStatus(db *dbb.Queries, ctx context.Context, wayUuid uu
...
-			CreatedAt:   dbMentor.CreatedAt.Format(util.DEFAULT_STRING_LAYOUT),
-			ImageUrl:    dbMentor.ImageUrl,
-			IsMentor:    dbMentor.IsMentor,
+			Uuid:        mentorMap["uuid"].(string),
+			Name:        mentorMap["name"].(string),
+			Email:       mentorMap["email"].(string),
+			Description: mentorMap["description"].(string),
+			CreatedAt:   createdAt.Format(util.DEFAULT_STRING_LAYOUT),
+			ImageUrl:    mentorMap["image_url"].(string),
+			IsMentor:    mentorMap["is_mentor"].(bool),
```
tritonJS826, line 335: It's very dangerous, looks like we have interface{} here. This means it will brings errors in run time unpredictable for any changes for user response here. Could we solve it somehow?

tritonJS826 (changes requested):

Issue #1145, opened 2024-06-20
All ways in a table view: Status completed doesn’t work


evgenia7p:
Preconditions
All ways page opened in a table view

Steps
1.	Click the Status select and set Completed from the dropdown menu
2.	Check the Status column 

Expected result
Only completed ways are displayed

Actual result
Abandoned and ways in progress are demonstrated as completed ways  

![Image](https://github.com/tritonJS826/masters-way/assets/163824877/44a23cd3-2071-47df-a4fd-52553648b243)

vdruda:
Preconditions
All ways page opened in a table view

Steps

Click the Status select and set Completed from the dropdown menu
Check the Status column
Expected result
Only completed ways are displayed

Actual result
Abandoned and ways in progress are demonstrated as completed ways

![Image](https://github.com/tritonJS826/masters-way/assets/141385609/208cdbab-67b7-4b4d-aa4c-eb64d2027507)

vdruda:
Done

AKiroi:
![Image](https://github.com/tritonJS826/masters-way/assets/112805668/814db318-b819-49c9-b591-78d873590de5)

vdruda:
![2](https://github.com/tritonJS826/masters-way/assets/141385609/908fb3aa-ff88-4a95-baa0-498f1ff0757a)

Done

PR #1185, opened 2024-06-24
fix: #1145 fixed status completed in AllWaysPage

AKiroi:
image


mw-webapp/src/dataAccessLogic/DTOToPreviewConverter/wayPlainDTOToWayPreview.ts
```diff
@@ -9,7 +9,7 @@ import {WayPreview} from "src/model/businessModelPreview/WayPreview";
  */
 export const wayPlainDTOToWayPreview = (wayDTO: SchemasWayPlainResponse): WayPreview => {
   const status = getWayStatus({
-    status: wayDTO.isCompleted ? "Completed" : null,
+    status: wayDTO.isCompleted ? "completed" : null,
```
Ekaterina1994, line 12: let's don't use just string 

we have enum WayStatus.completed

mw-webapp/src/dataAccessLogic/DTOToPreviewConverter/wayPlainDTOToWayPreview.ts
Ekaterina1994, line 1: Please check do we use string "completed" or other statuses anywhere in the code/ Let's update it too) And check do it work correctly)

Ekaterina1994 (changes requested):

mw-webapp/src/dataAccessLogic/DTOToPreviewConverter/wayPlainDTOToWayPreview.ts
AKiroi: done

mw-webapp/src/dataAccessLogic/DTOToPreviewConverter/wayPlainDTOToWayPreview.ts
AKiroi: fixed

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #1159, opened 2024-06-21
add migration script (base data) for testing puposes (insert related data in all tables + reset)


tritonJS826:
Add endpoint for dev environment to trigger this script and reset data in the database (for test purposes)
Notify Ekaterina (QA team) after implementation

tritonJS826:
done

PR #1203, opened 2024-06-27
feat: #1159 add migration script

F1zm0n:
should be good, if you have any notes about the structure feel free


mw-server/db/scripts/testdata/testdata.go
```diff
@@ -0,0 +1,188 @@
...
+	_ "github.com/golang-migrate/migrate/v4/source/file"
+	_ "github.com/lib/pq"
+	db "mwserver/db/sqlc"
+	"mwserver/util"
+	"net/http"
+	"strings"
+)
+
+var testData = `
+INSERT INTO "users" ("uuid", "name", "email", "description", "image_url", "is_mentor", "firebase_id")
```
tritonJS826, line 15: why not to move it to sql?

mw-server/db/scripts/testdata/testdata.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Adds test data
+// @Summary adds test data
+// @Description adds test data for testing purposes
+// @Tags test
+// @Accept json
+// @Produce json
+// @Success 200
+// @Router /test/data [post]
+func AddTestData(db *sql.DB) gin.HandlerFunc {
```
tritonJS826, line 151: we not to use controllers for such logic?

mw-server/db/scripts/testdata/testdata.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Resets all data
+// @Summary resets all data
+// @Description resets all data in the database for testing purposes
+// @Tags test
+// @Accept json
+// @Produce json
+// @Success 200
+// @Router /test/data [delete]
+func ResetData(q *db.Queries) gin.HandlerFunc {
```
tritonJS826, line 167: one public endpoint with  post method and no parameters which calls resetDb, and responsible for cleaning and data population should be enough for now

mw-server/db/scripts/testdata/testdata.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+
+// Resets and adds data
+// @Summary resets all data and adds test data
+// @Description resets all data and adds test data  for testing purposes
+// @Tags test
+// @Accept json
+// @Produce json
+// @Success 200
```
tritonJS826, line 180: wrong swagger schema - should return json as I see in the code

mw-server/db/scripts/testdata/testdata.go
F1zm0n: I wouldn't do that

mw-server/db/scripts/testdata/testdata.go
F1zm0n: Overhead

mw-server/db/scripts/testdata/testdata.go
tritonJS826: Create sql file for this ticket i\and convert it with sqlc generate to golang code. In this case it will be much more safe

With you approach you will be able to check this script only in runtime, string is not the best format for commands :)

mw-server/db/scripts/testdata/testdata.go
tritonJS826: we have conventions - all project for now use contollers. It is not db script it's just hadnlers for some endpoints

tritonJS826:
If you have any question ask any time me or @Moon1it

PR #1205, opened 2024-06-27
feat: #1118 edit assert and dataCy attribute for NoAuth_AllWays_SelectTableView test

EkaterinaSokolova-V:


mw-webapp/src/component/table/Table.tsx
```diff
@@ -29,6 +29,17 @@ interface TableData<T> {
...
+ * Data attributes for cypress testing
+ */
+interface dataCyTableContent {
+
+  /**
+   * Data attribute for cypress testing
+   */
+  dataCyTh?: string;
+}
+
```
tritonJS826, line 42: why do you need new interface for just one stiring value

EkaterinaSokolova-V: For other tests with table i'll need more attributes. So i created content interface. Leave it as it was? and then, if necessary, redo it?

EkaterinaSokolova-V: @tritonJS826, you're right, all other elements on the page are other components and there is no point in making an interface with attributes for the table.

tritonJS826 (approved):

Issue #989, opened 2024-06-04
Accessibility: theme switch button has no value text on Home page


pp2amd6:
## Description

Theme switch button has no value text on Home page. Accordingly to [WCAG](https://www.w3.org/TR/WCAG22/#non-text-content) descriptive text must be presented to screen reader users to indicate the function of the button.
![image](https://github.com/tritonJS826/masters-way/assets/4976837/27f2392d-fd10-479f-8782-e5e9f8529003)

## Steps to reproduce

1. Open Home page
2. View HTML code for theme switch button element

## Actual result

Theme switch button has no value text
```html
<button class="_iconWrapper_c6yhs_1 _themeSwitcher_15ikq_86" style="">...</button>
```

## Expected result

Theme switch button has value text

Ekaterina1994:
Need to add aria-label attribute instead of text value for icon button. Also need to research do we need to use this attribute to all buttons (where we use text value)? or only for icon buttons

pp2amd6:
> Also need to research do we need to use this attribute to all buttons (where we use text value)? or only for icon buttons

Only for icons because screen readers can read textContent for button tag.

> The button has an accessible label. By default, the accessible name is computed from any text content inside the button element. However, it can also be provided with [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) or [aria-label](https://w3c.github.io/aria/#aria-label).

Ekaterina1994:
Done

PR #1207, opened 2024-06-28
feat: #989 add accessibility in buttons on Home page [Pavel Kladkevich]

kladweb:
Add aria-label to theme switch button and sidebar button.
The buttons that have text value do not require aria-label.

### DIFF
--- a/mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
+++ b/mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
@@ -1,5 +1,6 @@
 import {ForwardedRef, forwardRef, PropsWithChildren} from "react";
 import {Trigger as DialogTrigger} from "@radix-ui/react-dialog";
+import {headerAccessIds} from "cypress/accessIds/headerAccessIds";
 
 /**
  * SidebarTrigger props
@@ -25,6 +26,7 @@ export const SidebarTrigger = forwardRef((props: PropsWithChildren<SidebarTrigge
       <div
         ref={ref}
         role="button"
+        aria-label={headerAccessIds.burgerMenu}
       >
         {props.children}
       </div>
--- a/mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
+++ b/mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
@@ -1,6 +1,5 @@
 import {ForwardedRef, forwardRef, PropsWithChildren} from "react";
 import {Trigger as DialogTrigger} from "@radix-ui/react-dialog";
-import {headerAccessIds} from "cypress/accessIds/headerAccessIds";
 
 /**
  * SidebarTrigger props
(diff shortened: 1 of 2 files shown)


mw-webapp/src/component/themeSwitcher/ThemeSwitcher.tsx
```diff
@@ -78,6 +78,7 @@ export const ThemeSwitcher = (props: ThemeSwitcherProps) => {
         className={clsx(styles.iconWrapper, props.className)}
         onClick={onChangeTheme}
         data-cy={props.dataCy}
+        aria-label={props.dataCy}
```
tritonJS826, line 81: We will probably remove data-attriutes for testing purposes from production some day

tritonJS826 (changes requested):

mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
```diff
@@ -25,6 +26,7 @@ export const SidebarTrigger = forwardRef((props: PropsWithChildren<SidebarTrigge
       <div
         ref={ref}
         role="button"
```
Ekaterina1994, line 28: 1. Task is not about adding aria-label to the Sidebar Trigger
2. Please read more about aria-label. Need research which tet should be in the aria and which tags couldn't include aria

For example, https://doka.guide/a11y/aria-label/


Please write  other oficial documentation about aria

Ekaterina1994 (changes requested):

mw-webapp/src/component/sidebar/SidebarTrigger/SidebarTrigger.tsx
kladweb: I think I should have added the aria tag only to the theme toggle button. The text value should match the language of the application.
Also I need to research and write for which elements we should add arria label or text values for accessibility.
Am I thinking in the right direction ?

Ekaterina1994: > I think I should have added the aria tag only to the theme toggle button.
Yes

Ekaterina1994: > The text value should match the language of the application.

Need research. It's a task to read  in the docs which value  there should be

Ekaterina1994: > Also I need to research and write for which elements we should add arria label or text values for accessibility.

Research need all time to check issues (maybe it's not a bug or bug is deprecated and etc.

kladweb: My research: 
https://docs.google.com/document/d/1pBUqgu7fRuseeu_52bSyRhO3oeVFuPvZ7gwjV_h8Wvw/edit?usp=sharing

Ekaterina1994 (approved):

tritonJS826 (approved):

mw-webapp/src/component/themeSwitcher/ThemeSwitcher.tsx
kladweb: Answered in dialogue with Ekaterina

PR #1235, opened 2024-07-01
feat: #1104 Align design for User Card component

algrinkevich:

### DIFF
--- a/mw-webapp/src/component/userCard/UserCard.tsx
+++ b/mw-webapp/src/component/userCard/UserCard.tsx
@@ -70,54 +70,66 @@ export const UserCard = observer((props: UserCardProps) => {
             <Avatar
               alt={props.userPreview.name}
               src={props.userPreview.imageUrl}
-              size={AvatarSize.BIG}
+              size={AvatarSize.MEDIUM}
             />
-            <VerticalContainer className={styles.nameGroup}>
-              <Title
-                text={props.userPreview.name}
-                level={HeadingLevel.h3}
-                className={styles.title}
-                placeholder=""
-              />
-              <p className={styles.mail}>
-                {props.userPreview.email}
-              </p>
-            </VerticalContainer>
-            <VerticalContainer className={styles.likes}>
+            <Title
+              text={props.userPreview.name}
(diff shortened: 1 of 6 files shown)


mw-webapp/src/component/userCard/UserCard.tsx
```diff
@@ -109,15 +109,25 @@ export const UserCard = observer((props: UserCardProps) => {
...
+            <span>
+              {`${LanguageService.allUsers.userCard.createdAt[language]}`}
+            </span>
+            <span className={styles.createdAt}>
+              {DateUtils.getShortISODotSplitted(props.userPreview.createdAt)}
+            </span>
+          </HorizontalContainer>
+          <HorizontalContainer className={styles.ways}>
+            <span>
+              {`${LanguageService.allUsers.userCard.ownWays[language]} (${props.userPreview.ownWays})`}
```
Ekaterina1994, line 122: Is it possible don't use span wrappers for each element?

mw-webapp/src/globalStore/ThemeStore.ts
```diff
@@ -53,6 +53,10 @@ const themedVariables: Record<string, Record<Theme, string>> = {
     [Theme.DARK]: "#8F8F8F",
     [Theme.LIGHT]: "#A8A1DD",
   },
+  secondaryCardsBgColor: {
```
Ekaterina1994, line 56: do we have primaryCradsBgColor? If not then name is not actual. You can use name for example additionalInfoCardsBgColor

algrinkevich: No, we don't have the primary. I renamed to  `additionalInfoCardBgColor`

mw-webapp/src/component/userCard/UserCard.tsx
algrinkevich: If I remove the wrappers, there will be a small distance between "ways", not the same as in the design.

Design:
image

Without wrappers:
image

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #785, opened 2024-04-29
add flag-marks on user card if it is a mentor


tritonJS826:
- [ ] Add flag
- [ ] Align design for user card

Ekaterina1994:
Not clear. Please add links on figma or add comment what should to do developers)

tritonJS826:
mentor flag - green

tritonJS826:
Need token for green color

vikpak89:
creating token for bookmarks 

https://www.figma.com/design/ukyPUFcecdKzeBcdTHChYk/%D0%9F%D1%80%D0%BE%D0%B5%D0%BA%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5?node-id=2596-38020&t=9n8J9WzzyiowONLs-1

Ekaterina1994:
Blocked by #926

tritonJS826:
just isMentor green flag will be enough for now

Ekaterina1994:
UNBLOCKED

Ekaterina1994:
Done

evgenia7p:
Checked

PR #1263, opened 2024-07-05
feat: #785 add menor flag

AKiroi:
- added mentor flag 
image

### DIFF
--- a/mw-webapp/src/component/userCard/UserCard.module.scss
+++ b/mw-webapp/src/component/userCard/UserCard.module.scss
@@ -2,12 +2,14 @@
 
 $userCardMainHeight: 140px;
 $userCardAdditionalHeight: 82px;
+$flagMenorRightOffset: 24px;
 
 .cardLink {
   text-decoration: none;
 }
 
 .userCardContainer {
+  position: relative;
   overflow: hidden;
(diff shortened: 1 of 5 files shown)


mw-webapp/src/component/userCard/UserCard.module.scss
```diff
@@ -83,3 +85,11 @@ $userCardAdditionalHeight: 82px;
 .ways {
   flex-wrap: wrap;
 }
+
+.flagMentor {
+  position: absolute;
+  top: 0;
+  right: $flagMenorRightOffset;
+  fill: var(--bookmarksIconColor);
+  stroke: none
```
Ekaterina1994, line 94:
```suggestion
  stroke: none;
```

Ekaterina1994, line 94: Why stroke is 0?

AKiroi: in figma maket we don't have stroke, only fill 
image

AKiroi: done

Ekaterina1994 (approved):

tritonJS826 (approved):

Issue #1190, opened 2024-06-25
User personal data: Avatar blends into the background


evgenia7p:
Preconditions
User is authorized. Personal area page is open

Steps
Check the avatar of the user in the top left corner

Expected result
Avatar color in a circle form is different from background color

Actual result
Avatar blends into the background

![Image](https://github.com/tritonJS826/masters-way/assets/163824877/12bb28bc-9915-4ae7-8f6b-cd497c19cf37)

Ekaterina1994:
We don't have avatar in components in the maket in figma. It's not a business requirement that background should not repeat color of layer.  Also after updating login new users have colored avatars (as in google account). So I'm not sure that we need customize old user's abatars. If need -   Move issue to design columns

vikpak89:
created components for Avatars of different sizes
https://www.figma.com/design/ukyPUFcecdKzeBcdTHChYk/%D0%9F%D1%80%D0%BE%D0%B5%D0%BA%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5?node-id=4190-9892&t=HlcH20UrrMgqhBaJ-1



![Image](https://github.com/tritonJS826/masters-way/assets/167992048/e34ce37f-8623-4eb0-a035-ee864f987c6f)

timoxa20:
Add a frame for the avatar so that it is the same on each page and does not match the background color

vikpak89:
added frames to avatars
https://www.figma.com/design/ukyPUFcecdKzeBcdTHChYk/%D0%9F%D1%80%D0%BE%D0%B5%D0%BA%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5?node-id=4190-9892&t=egAB04ZUjesdvhYB-1

Ekaterina1994:
done

evgenia7p:
Checked

PR #1266, opened 2024-07-06
fix: #1190 Avatar blends into the background

timoxa20:
Fix Changed the background color of the Avatar component, now it does not merge with the background

### DIFF
--- a/mw-webapp/src/component/avatar/Avatar.module.scss
+++ b/mw-webapp/src/component/avatar/Avatar.module.scss
@@ -28,7 +28,7 @@ $avatarLargeSize: 120px;
   height: 100%;
   align-items: center;
   justify-content: center;
-  background-color: var(--secondaryBgColor);
+  background-color: var(--generalPrimaryColor);
   color: var(--primaryTextColor);
 }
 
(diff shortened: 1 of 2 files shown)


mw-webapp/src/component/avatar/Avatar.tsx
```diff
@@ -69,26 +70,51 @@ const getInitials = (name: string): string => {
...
-    >
-      <AvatarRadix.Image
-        className={styles.AvatarImage}
-        src={props.src ?? ""}
-        alt={initials}
-      />
-      <AvatarRadix.Fallback
-        className={styles.AvatarFallback}
-        delayMs={600}
+  if (location.pathname === "/") {
```
tritonJS826, line 75: super dangerous, avatar as component should not change it behavior in depends on the location

tritonJS826 (changes requested):

mw-webapp/src/component/avatar/Avatar.tsx
Ekaterina1994: Agree, discused about it in telegram

mw-webapp/src/component/avatar/Avatar.module.scss
```diff
@@ -13,6 +13,7 @@ $avatarLargeSize: 120px;
   border-radius: $tertiaryBorderRadius;
   user-select: none;
   vertical-align: middle;
+  border: 1px solid var(--primaryStrokeColor)
```
Ekaterina1994, line 16: global var instead 1px

timoxa20: Fix border number 1px на global varieibls

Ekaterina1994 (approved):

Issue #1250, opened 2024-07-03
Integration with AI Gemini: Add endpoints, create a client, and handle incoming prompts


Moon1it:
- [x] Define API Endpoints
- [x] Implement Client Library
- [x] Handle Incoming Prompts

n144astya:
Done

PR #1275, opened 2024-07-11
feat: #1250 Refactored gemini integration, renamed AI to Gemini, crea…

Moon1it:
…ted gemini client in the root of project. Made prompt and got JSON from gemini.

https://github.com/tritonJS826/masters-way/issues/1250


mw-server/main.go
```diff
@@ -129,6 +134,11 @@ func init() {
 
 	fmt.Println("PostgreSql connected successfully...")
 
+	geminiClient, err = genai.NewClient(ctx, option.WithAPIKey(config.Env.GeminiApiKey))
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+	}
```
tritonJS826, line 140: let's checl better for env type (prod vs dev instead of checking variables itself)

Moon1it: Yes, I will add env type check, although there is no variable check or any other check

Moon1it: Done

Issue #726, opened 2024-04-21
[Suggestion] Mentor: show the ways that await the actions of the mentee.


pp2amd6:
Now, after a user has submitted an application to become a mentor, he only sees a notification in the path from which the application was sent. Information about submitted applications is not displayed to the user anywhere else.
![image](https://github.com/tritonJS826/masters-way/assets/4976837/ff9aad8a-1d46-4596-8535-2d865d1e83a6)

I suggest showing the path for which the application was sent in the user's (mentor's) profile.
![image](https://github.com/tritonJS826/masters-way/assets/4976837/e51ac22b-2561-4b01-938a-80aca39a2da8)

There is a button to cancel the request
![image](https://github.com/tritonJS826/masters-way/assets/4976837/441cbb12-0597-47bf-ab23-e963fe4d8d2d)

Also add some kind of indication to the path counter so the user can pay attention.
It could be anything:
asterisk symbol,
![image](https://github.com/tritonJS826/masters-way/assets/4976837/aaf695c2-4fa5-4e44-8c8c-b5426dc6024a)

counter, like mobile apps or something else
![image](https://github.com/tritonJS826/masters-way/assets/4976837/962211f0-aa0f-4d50-94d7-1d0bcbcbc9bb)

This functionality will allow you to better understand what paths still await action from the mentee.

tritonJS826:
i think it will be better to add additional optional collections "requests". Will be added in future

tritonJS826:
Yeah it will, but let's close it after implementation

PR #1296, opened 2024-07-17
feat: #726 add new controller, endpoint after rebase

F1zm0n:


tritonJS826 (changes requested):
Lets modify existent one endpoint (getUserById) consistently instead of adding another one

Issue #1294, opened 2024-07-17
Add ability to remove ways from composite from parent


Ekaterina1994:
It is required for cases when you added public way to composite and aftrer that this way become private (I already faced with such case).

Ekaterina1994:
Done

PR #1298, opened 2024-07-17
feat: #1294 Add ability to delete way from composite way page

Ekaterina1994:


mw-server/controllers/way.controller.go
```diff
@@ -53,6 +53,7 @@ func (cc *WayController) CreateWay(ctx *gin.Context) {
 	util.HandleErrorGin(ctx, err)
 
 	now := time.Now()
+	var copiedFromWayUuid pgtype.UUID
```
Ekaterina1994, line 56:
```suggestion
```

tritonJS826 (approved):

Issue #1115, opened 2024-06-19
Day Reports: A word ‘tag’ is presented instead of ‘label’ in some cases


evgenia7p:
Preconditions
User is authorized. User's way is created and opened. Jobs done labels exist. Day report with labels exists

Steps
1.	Click on "Adjust job tags" in the Reports section
2.	Hover over bin icon in opened dialog window
3.	Close dialog window
4.	Click on any label in the report
5.	Hover over marked and unmarked labels
6.	Check all languages

Expected result
Title of the button is “Adjust job labels” (not ‘tegs’). “Label” is instead of “tag” in the dialog window for adusting job labels (button “Add label”, tooltip “Delete from jobs labels”, text “Are you sure you want to remove ‘…’ from job labels?”). The same in the tooltips “Click to add label” and “Click to remove label” in reports. It is the same for all languages.

Actual result
A word ‘tag’ is presented instead of ‘label’ in six cases

![Image](https://github.com/tritonJS826/masters-way/assets/163824877/0bed881c-4541-4dc3-b51f-02ad7a3f83c4)
![Image](https://github.com/tritonJS826/masters-way/assets/163824877/a03284c9-7309-432a-9bf2-45f3499989a8)
![Image](https://github.com/tritonJS826/masters-way/assets/163824877/15905e71-8653-4865-b756-4da8fa4b7261)
![Image](https://github.com/tritonJS826/masters-way/assets/163824877/8206134c-5e30-4128-9f21-d04ace5c95cf)
![Image](https://github.com/tritonJS826/masters-way/assets/163824877/6c7836c1-8d8c-478b-a8ad-fe926221b25b)

Ekaterina1994:
Done

evgenia7p:
Tooltips «Delete from job labels», «Click to add label» и «Click to remove label» are gone. Please, return them. Also button “Add label” should be instead “Add labelS” (single number instead of plural)

timoxa20:
I fixed it. You can check if there are doubts about: Also the button “Add label” should be instead of “Add labelS" (single number instead of plural). Because I found the Add label in only one place

evgenia7p:
Checked

evgenia7p:
A word ‘tag’ is still presented instead of ‘label’ in Russian and Ukrainian versions

Ekaterina1994:
Не нашла где используется слово тэг вместо лейбл. Но заменила слова "метка" на "лейбл"

n144astya:
Done

PR #1326, opened 2024-07-21
fix: #1115 Fix day report

timoxa20:
fixed Added a value for the tooltip so that it would not hide behind the modal window


tritonJS826 (approved):

mw-webapp/src/styles/_variables.scss
```diff
@@ -38,8 +38,8 @@ $tertiaryTransitionDuration: 500ms;
 
 // z-index
 $sidebarZIndex: 100;
+$tooltipZIndex: 121;
```
Ekaterina1994, line 41:
```suggestion
$tooltipZIndex: 130;
```

Ekaterina1994 (approved):

Issue #1307, opened 2024-07-18
Create backend for Master's Way chat service (mw-chat)


Moon1it:
## Description
Develop the backend for the chat service.

- [ ] Creating endpoints
- [ ] Integration with the General service (Users)
- [ ] Structuring the application (controllers, services, repositories)
- [ ] Designing and implementing the database

tritonJS826:
private chats done

PR #1340, opened 2024-07-23
feat: #1307 4 out of 11 endpoints were implemented for GroupRooms.

Moon1it:
Added controllers:
- HandleGetGroupRooms
- HandleGetGroupRoomById
- HandleCreateGroupRoom
- HandleUpdateGroupRoom

Added services:
- GetGroupRoomsPreview
- GetPopulatedGroupRoom
- CreateGroupRoom
- BlockOrUnblockGroupRoom


mw-chat/internal/services/group_rooms.service.go
```diff
@@ -1,56 +1,168 @@
...
-	// groupRoomId := ctx.Param("groupRoomId")
-	// fmt.Println(groupRoomId)
+func (groupRoomsService *GroupRoomsService) CreateGroupRoom(ctx context.Context, roomName string, userUUID uuid.UUID) (*schemas.RoomPopulatedResponse, error) {
+	now := time.Now()
+
+	tx, err := groupRoomsService.pool.Begin(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer tx.Rollback(ctx)
```
tritonJS826, line 97: will it be executed on all scenarios?

tritonJS826 (approved):

mw-chat/internal/services/group_rooms.service.go
Moon1it: Yes, I found it in sql doc. It standard scenario. When we do it after commit, it's alright.

PR #1406, opened 2024-08-02
fix: #1307 Added tests for mw-chat-bff and mw-chat. Added field image…

Moon1it:
…URL to Rooms. Read messages when returning the room by id.

- Added tests for mw-chat-bff and mw-chat.
- Added field imageURL to RoomPopulatedResponse and RoomPreviewResponse.
- Read messages when returning the room by id.
- Fixed bug: room could be created for non-existent user. Now we check if the user exists in general service. If the user doesn't exist general service returns error.


mw-chat-bff/internal/services/rooms.service.go
Moon1it, line 1: To receive an error from other servers, we need to take it from response (second returned param) text of error, because err (third returned param) involves only status code.

mw-chat/internal/server/server_test.go
```diff
@@ -98,7 +98,7 @@ func sendMessage(t *testing.T, roomID, userID, message string) {
 	var roomPopulatedResponse schemas.MessageResponse
 	MakeTestRequestWithJWT(t,
 		http.MethodPost,
-		newConfig.ChatBaseUrl+"/chat/rooms/"+roomID+"/messages",
+		newConfig.ChatBaseURL+"/chat/rooms/"+roomID+"/messages",
```
tritonJS826, line 101: getChatCreateMessageUrl(baseUrl, roomId) string

tritonJS826 (approved):

mw-chat/internal/db/query/message_status.query.sql
```diff
@@ -7,3 +7,12 @@ WITH room_users AS (
 INSERT INTO message_status (message_uuid, receiver_uuid, is_read)
 SELECT @message_uuid, user_uuid, false
 FROM room_users;
+
+-- name: ReadMessages :exec
```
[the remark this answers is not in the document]
Moon1it: Done

